PR #57, opened 2025-01-18
feat: Add waveshare panel driver

jniebuhr:


shark526:
hi @jniebuhr, I've noticed waveshare driver in the source code, since I have a waveshare esp32-s3-touch-2.1 screen, I'm not sure if it compatible for the gaggimate display project.
If it also work, could you please point out which config should I change to apply to the 2.1 waveshare screen?   I tried to direct flash the display code to waveshare screen, but most of time it not light up, only a few time it works.

Thank you!

PR #106, opened 2025-02-14
Feature/volume prediction

smury:
Prediction of brewDelay and grindDelay based on a linear fit to the last PREDICTIVE_TIME = 2s of measured volumes (weights) and times recorded in measurements and measurementTimes

Delays are updated 2s after the processes become inactive (!isActive), after which they are deactivated (deactivate()) by the controller.


jniebuhr:
can you put the volumepersecond and similar methods into a separate class like in our discord discussion? should fix the duplication issue

smury:
I don't recall such a discussion but I put the linear fit calculation into a new header file. The linear fit may be useful later when pressure control comes. I could also add other sciency functions e.g. interpolate the flow-pressure curve to predict the flow at any given pressure and use the pressure/flow of the current shot to predicted the current weight in the cup.

I actually had a brain-failure with the code duplication warning - I thought it was referring to the similarity between my changes and the pre-existing code!

PR #175, opened 2025-04-07
feat: Introduce profiles

### LINKED ISSUE #7: Allow saved profiles with pump on/off

jniebuhr:
Fixes #7, #8, #9


schema/profile.json
```diff
@@ -0,0 +1,136 @@
...
+    },
+    "description": {
+      "type": "string"
+    },
+    "favorite": {
+      "type": "boolean"
+    },
+    "selected": {
+      "type": "boolean"
+    },
```
hanstf, line 32: Hello, as I wanted to try to convert the gaggiuino profile to this schema, for example this https://github.com/Zer0-bit/gaggiuino/blob/community/profiles/Tea%20Mofos%20Reloaded.json#L120
Seems I could not see the temperature field in this schema, how do I set the temperature target in here ?

PR #230, opened 2025-06-04
Adjust UI pressure scale gauge

0x64c:
Reads pressure scale value stored in settings on initialization and applies it to the scale of the gauge. Requires a restart after setting is changed.


0x64c:
not 100% certain it's bug-free for all sane values, but it seems to work okay with my 12-bar sensor.

0x64c:
To be thorough, I would suggest testing with different sensors.

Issue #130, opened 2025-03-09
[Feature Request] Options to display temp in Fahrenheit and adjust time on clock


lylelotspeich:
### Describe the feature

Is there any way we can add options in the settings to change from Celcius to Fahrenheit and add the option to change the clock to 12-hour and whatever the local time is?

I also think it would be nice to have an option to have the standby screen just be the clock displaying the local time, with the Gaggimate logo smaller on the top or bottom of the screen - but this is purely a cosmetic opinion.

### Why is this feature useful?

In the US most people use Fahrenheit, and the clock is unadjustable to my knowledge, takes away from the look of the screen when it is showing the wrong time.

### How complex is this feature?

Low - Simple enhancement

jniebuhr:
Time adjustment was finished, Fahrenheit adjustment was moved to #244

PR #246, opened 2025-06-14
Add Clock format to settings and UI

silverfoxy:
This PR adds a new settings toggle to choose between 24h and 12h clock format mentioned in https://github.com/jniebuhr/gaggimate/issues/130

Mainly getting myself familiar with the code base so I started with some smaller task. I tested the change both using a mock for the API using vite-plugin-mock (which I can add in a separate PR if you're interested to have such tests without relying on the controller to respond to API calls). I also flashed it on my device and tested it that way. 

**Web UI**

**Display**
<p float>

</p>


jniebuhr:
Thanks for the contribution. Yeah that mock would also be appreciated!

PR #330, opened 2025-07-27
Add pump model settings to WebUI

spsancti:
Adds WebUI configuration for pump model coefficients (_Q0, _Q1) used in PressureController for accurate pump flow modeling, following the same pattern as PID parameters.

**Changes**
Settings: Added getPumpModelCoeffs()/setPumpModelCoeffs() with persistent storage
WebUI: Added form field for pump model coefficients in Settings page
BLE: Added pump model coefficients characteristic and callback registration
Controller: Added setPumpModelCoeffs() method and integration with PressureController

**Usage**
Configure pump model coefficients via Settings → Machine settings → "Pump Model Coefficients (Q0, Q1)" field (default: "10.79, -0.5854").

## Summary by CodeRabbit

* **New Features**
  * Configure pump flow coefficients (2-value linear or 4-value polynomial) via Settings and Web UI.
  * New BLE characteristic to remotely update pump flow coefficients.
  * Pump flow coefficients are persisted with machine settings.

* **Improvements**
  * Coefficients are transmitted to compatible devices on update and at initial connection.
  * Pump flow model upgraded from linear to a third-order polynomial for improved accuracy.


nayrod485:
For those who are not versed into calculating the slope and offset of their pump flow rate, it'll be easier to input flow rates measured at two pressure points rather than computing the affine model. 

 For those who are mathematically incline to make a linear regression over multiple pressure point, they would be capable of computing the flow rate at two pressure point from their linear regression function. 

The former solution is not restrictive and more user friendly. 

Additionally more data need to be collected to assess that the linear model is good enought and may need to switch over third order polynomial. This is to think about future evolution of that input field.

spsancti:
Okay, I'll revert. 
I think it will be prone to errors because it's quite finicky to achieve an exact and stable pressure with these valves. However, I agree that it's more user-friendly to input two directly measured values simply. 
Do we need to support the pump flow poly, or is linear model good enough?

lib/GaggiMateController/src/peripherals/DimmedPump.cpp
```diff
@@ -79,3 +79,11 @@ void DimmedPump::setPressureTarget(float targetPressure, float flowLimit) {
 }
 
 void DimmedPump::setValveState(bool open) { _valveStatus = open; }
+
+void DimmedPump::setPumpFlowCoeff(float oneBarFlow, float nineBarFlow) {
```
nayrod485, line 81: One could simplify the pressureController to have only one setPumpCoeff(a,b,c,d) method with polynomial coefficient computed by the Dimmer class directly

lib/GaggiMateController/src/GaggiMateController.cpp
```diff
@@ -80,6 +80,17 @@ void GaggiMateController::setup() {
...
     _ble.registerAltControlCallback([this](bool state) { this->alt->set(state); });
     _ble.registerPidControlCallback([this](float Kp, float Ki, float Kd) { this->heater->setTunings(Kp, Ki, Kd); });
+    _ble.registerPumpModelCoeffsCallback([this](float a, float b, float c, float d) { 
+        if (_config.capabilites.dimming) {
+            auto dimmedPump = static_cast<DimmedPump *>(pump);
+            // Check if this is a flow measurement call (a and b are flow measurements, c and d are 0)
+            if (c == 0.0f && d == 0.0f) {
+                dimmedPump->setPumpFlowCoeff(a, b); // a = oneBarFlow, b = nineBarFlow
+            } else {
+                dimmedPump->setPumpFlowPolyCoeffs(a, b, c, d); // a, b, c, d are polynomial coefficients
```
nayrod485, line 109: One could keep this function only in the PressureController class to setup coefficients

nayrod485 (approved):
The proposed alternative is compatible with the pressure controller logic

lib/GaggiMateController/src/GaggiMateController.cpp
spsancti: I'd like to keep it separate for the clarity
I made it separate so the extra two coefficients are NaN in flow mode and can be anything in the poly mode

spsancti:
@coderabbitai review

PR #339, opened 2025-07-31
Add Daisy Ui and Convert Home Page 

ManfredRichthofen:
This Change adds
- Daisy UI with four themes that are not even close to final.
- Restyles the process controls card as suggested in discord (need to test as im on 1.5.1)


ManfredRichthofen:
I marked this ready for review so i can see what coderabbit says about it.

PR #343, opened 2025-08-04
Feat/hw scales

ericyanush:
A WIP implementation of the Hardware Scales mod. 

## Overview

What's included:

-  [x] Integration of two HX711 modules with shared clock line (tested on GM Pro v1)
  -  [x] Taring of scales
  - [x] Calibrating individual load cells
  - [x] Basic reading averaging
  - [ ] HX711 power down   
- [x] Integration of Hardware scale into display firmware
  - [x] HardwareScalePlugin added
  - [ ] Combine with BLEScalePlugin / adjust BLEScalePlugin
  - [x] Hardware scale calibration factors settings
  - [ ] WebUI for performing scale calibration
  - [ ] Add tare button to WebUI dashboard?
  - [ ] Add live weight display to brewing "start" screen?
  - [ ] Add tare button to brewing "start" screen?  

## Open questions / Feedback needed

Some notes, and places I'd like some feedback:

1. Feedback on changes to the WebUI would be especially appreciated; my experience with React is quite low 🙈
2. I added the display of the current weight to the status and chart views on the webui home page, mostly out of convenience for testing as the currently weight doesn't show anywhere else.
3. I have the new code for LED / ToF commented out of the controller startup, as their use of the 5-pin port interferes with the HX711's. The HX711 uses a two wire serial (SPI like) signalling protocol, which isn't compatible with the I2C that is being used on other expansion boards.
4. I'm currently piggybacking on the BLE volumetric measurement source. A third source could be added rather easily, but  for the sake of proving out the rest of implementation, I took the easy path forward. 
5. I currently have the BLEScalePlugin disabled in (commented out the register plugin call) in the display firmware. Right now the BLEScalePlugin makes the assumption that it is the only source of enabling and disabling volumetric overrides, and updates the status of the override in its runloop. I see two possible paths forward here:
  a. Combine the BLEScalePlugin and the HardwareScalePlugin into a single plugin that allows for only one type of scale to be active at a time.
  b. Change the type of the volumetric override from a bool to an enum that indicates what the source of the override is, then in the setVolumetricOverride method, check to  see if the source of the change request matches the currently configured one, something like:
```C++
enum VolumetricOverrideSource {
    NONE,
    BLUETOOTH,
    HARDWARE_SCALE,
};

void setVolumetricOverride(bool override, VolumetricOverrideSource source) {
    if (_override == source || _override == NONE) {
       _override = override ? source : NONE;
    }
}

```

## Testing

I've completed all my testing on a GCP E24 with GM Pro v1.

### Test setup:
- Hardware scale mod, using mostly standard parts found in the printables project (I had to narrow up the low profile drip tray adapter to fit the LP drip tray I purchased from Aliexpress).
- Hardware scales with two discrete "green" HX711 boards.
  - HX711 boards modified with a 4.7K/10K voltage divider on the DT lines to ensure that we don't expose the ESP32 on the controller board to more than ~3.4V. Alternatively a bi-directional MOSFET type level shifter could be used, but without easy access to a 3.3V source, the voltage divider is simpler to implement.

### Test notes:
- The predictive scale feature is rather aggressive in my testing, and I've had to lower its time interval to 200ms to not dramatically undershoot the target weights.

### DIFF
--- a/lib/GaggiMateController/src/ControllerConfig.h
+++ b/lib/GaggiMateController/src/ControllerConfig.h
@@ -8,6 +8,7 @@ struct Capabilities {
     bool ssrPump;
     bool ledControls;
     bool tof;
+    bool hwScale;
 };
 
 struct ControllerConfig {
--- a/lib/GaggiMateController/src/GaggiMateController.cpp
+++ b/lib/GaggiMateController/src/GaggiMateController.cpp
@@ -37,16 +37,33 @@ void GaggiMateController::setup() {
     this->brewBtn = new DigitalInput(_config.brewButtonPin, [this](const bool state) { _ble.sendBrewBtnState(state); });
     this->steamBtn = new DigitalInput(_config.steamButtonPin, [this](const bool state) { _ble.sendSteamBtnState(state); });
 
-    // 5-Pin peripheral port
-    Wire.begin(_config.scaleSdaPin, _config.scaleSclPin, 400000);
-    this->ledController = new LedController(&Wire);
-    this->distanceSensor = new DistanceSensor(&Wire, [this](int distance) { _ble.sendTofMeasurement(distance); });
-    if (this->ledController->isAvailable()) {
-        _config.capabilites.ledControls = true;
-        _config.capabilites.tof = true;
-        _ble.registerLedControlCallback(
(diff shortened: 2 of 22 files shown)


Shin-nn:
Hi, are there any future plans for this? I would love this feature and i guess i could find some time to work on it, if there is any chance it will be merged.

PR #357, opened 2025-08-14
Shot history account for post brewing drip

dragm83:
Fix for shot history not accounting for liquid still in transit after brew ends and recording an incorrect (premature) brew weight. This history will record for an additional 3 seconds OR until the weight stabilizes (<0.1g delta in 1s). 

## Summary by CodeRabbit

- New Features
  - Extended post-shot recording to capture stabilized weight/volume data, increased saved history entries from 3 to 6, and added per-entry notes persistence (save/load).

- Bug Fixes
  - Ignores invalid/spiking weight readings and skips malformed history lines.
  - Keeps history file open during extended recording to avoid premature closure.
  - Automatically discards very short recordings; history deletion removes associated note files.

- Refactor
  - Reworked recording lifecycle, stabilization logic, and finalization pathway for more reliable shot history.


spsancti:
Please replace all the try...catch with explicit error handling. In embedded builds, the exception mechanism is implementation-dependent and may not work as expected because there is nothing to throw an exception.

PR #362, opened 2025-08-15
Show BLE Scale Weight and Total Volumetric Target in WebUI

dragm83:
Show current Bluetooth weight reading in WebUI.
Automatically switch to Volumetric mode when a scale connects (if volumetric targets > 0).
Show total volumetric target from loaded profile even with multiple phases.

Desktop:
image

Mobile:
image

## Summary by CodeRabbit

* **New Features**
  * Live Bluetooth scale: shows current scale weight (0.1g) in two header locations when connected; displays configured weight target alongside live weight when a weight target is active.
  * Time/Weight toggle: switches based on whether a weight target is set (Time when not using weight target; Weight when set).

* **Chores**
  * Bluetooth connection status and weight added to status payload and persisted in machine state/history.
* **Bug Fixes**
  * Status now reports numeric volumetric targets consistently.


src/display/plugins/BLEScalePlugin.cpp
```diff
@@ -70,7 +70,19 @@ void BLEScalePlugin::loop() {
 }
 
 void BLEScalePlugin::update() {
-    controller->setVolumetricOverride(scale != nullptr && scale->isConnected());
+    bool currentConnectionState = scale != nullptr && scale->isConnected();
+    controller->setVolumetricOverride(currentConnectionState);
+    
+    // Auto-switch to weight mode when scale connects (if not already in weight mode)
```
jniebuhr: this is a user choice if he'd like to use the scale. the actual switching is already done 2 lines above

src/display/plugins/WebUIPlugin.cpp
```diff
@@ -70,8 +76,28 @@ void WebUIPlugin::loop() {
...
         doc["cd"] = controller->getSystemInfo().capabilities.dimming;
-        doc["bt"] = controller->isVolumetricAvailable() && controller->getSettings().isVolumetricTarget() ? 1 : 0;
+        
+        // Calculate total volumetric target weight from all phases
+        float totalVolumetricTarget = 0.0f;
+        Profile selectedProfile = controller->getProfileManager()->getSelectedProfile();
+        for (const auto &phase : selectedProfile.phases) {
+            if (phase.hasVolumetricTarget()) {
+                Target target = phase.getVolumetricTarget();
+                totalVolumetricTarget += target.value;
```
jniebuhr: this will result in the wrong value, this function on the profile is more suitable:

`    double getTotalVolume() const {
        double volume = 0.0;
        for (const auto &phase : phases) {
            if (phase.hasVolumetricTarget()) {
                volume = phase.getVolumetricTarget().value;
            }
        }
        return volume;
    }`

src/display/plugins/BLEScalePlugin.cpp
dragm83: Updated and removed

src/display/plugins/WebUIPlugin.cpp
dragm83: This is similar to the controller->getBrewVolume except I'm looking to see what the total weight is over all phases of a profile (since that's what the scale will reflect) - unless I misunderstood?

`    double getBrewVolume() const {
        double brewVolume = 0;
        for (const auto &phase : profile.phases) {
            if (phase.hasVolumetricTarget()) {
                Target target = phase.getVolumetricTarget();
                brewVolume = target.value;
            }
        }
        return brewVolume;
    }`

dragm83: misunderstood this before, fixed.

PR #379, opened 2025-08-22
Fix iPhone homescreen save icon

asvartsjo:
Add the apple-touch-icon link to ensure the correct icon displays when saving to the homescreen on an iPhone.


asvartsjo:
@jniebuhr Don't know if I can do anything to fix the deploy preview?

PR #384, opened 2025-08-22
Schedule Warmup/Brew

dragm83:
This adds the ability to schedule the GaggiMate to enter brew mode automatically one or more times a day. This effectively serves as an automatic 'warm up' scheduler as the machine will get up to temp based on the last selected profile. 

image

## Summary by CodeRabbit

* **New Features**
  * Auto Wakeup scheduling: device can automatically start warming/brewing from standby at configured times.
  * Settings UI & API: toggle to enable/disable Auto Wakeup and manage multiple times (add, edit, remove); schedule persists with sensible default "07:00".

* **Bug Fixes / Notes**
  * Prevents repeated activations within the same minute.
  * Duplicate UI block inserted in the settings page (non-functional).

### DIFF
--- a/src/display/core/Controller.cpp
+++ b/src/display/core/Controller.cpp
@@ -477,7 +477,6 @@ void Controller::lowerGrindTarget() {
 }
 
 void Controller::updateControl() {
-    unsigned long now = millis();
     float targetTemp = getTargetTemp();
     if (targetTemp > .0f) {
         targetTemp = targetTemp + static_cast<float>(settings.getTemperatureOffset());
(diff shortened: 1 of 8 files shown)


jniebuhr:
Can you rename these variables internally to auto wake up or something similar? Auto brew sounds like it actually starts brewing. Also please don’t use updateControl for the check, that method is reserved to update the controller

dragm83:
> Can you rename these variables internally to auto wake up or something similar? Auto brew sounds like it actually starts brewing. Also please don’t use updateControl for the check, that method is reserved to update the controller

I've migrated the wake up function to be its own plugin.

spsancti:
It would be cool to add a day of the week to the schedule, so workdays and weekends can have different times

PR #404, opened 2025-09-02
Visualizer.coffee API integration from Shot History

dragm83:
Post shot data directly to visualizer.coffee from shot history card

image

image

image


silverfoxy:
What do you think about storing the credentials on GM? I think it would have a better UX (when switching devices) and also allows for auto upload functionality.

dragm83:
> What do you think about storing the credentials on GM? I think it would have a better UX (when switching devices) and also allows for auto upload functionality.

In the current implementation, everything is handled by the front-end/browser so it's fairly light on the controller and doesn't have to use the ESP32s Http client library which isn't the nicest. Although I agree, it would be nice to handle it internally and have a toggle for 'auto upload shot'. 

However, since the current workflow lets users enter some notes after a shot completes, it also does make sense to keep it this way and have the user choose after the fact if they fill out the notes to post the shot. 

Would be good to push the changes on the visualizer.coffee end to support the notes data from Gaggimate.

PR #413, opened 2025-09-09
WebUI Optimizations

dragm83:
- Add Websocket heartbeat to drop stale clients
- Add delta updates to only send changed data rather than full payload every 500ms


dragm83:
Rewriting with a more streamlined approach.

PR #432, opened 2025-09-22
OTA/WS Stability Guards

dragm83:
Remove VLAs
Add null pointer checks
Limit buffer sizes

Suggest reducing automatic OTA update check frequency to once per week - update can be checked on webui ota save/refresh anyways.


jniebuhr:
Extracting the Web UI part as the rest is about to change

PR #441, opened 2025-09-30
Optimize/shot history + Storage indicator in system & updates

dragm83:
Replace CSV history format with a compact binary format written directly to SPIFFS to reduce CPU time and flash wear from frequent string allocations / formatting. 

Handle Incomplete / Interrupted Shots
If the device reboots or the shot is otherwise interrupted before finalization, `sampleCount` and `durationMs` remain 0 in the header. Recovery logic:
- Listing (`req:history:list`) infers `samples` and `duration` by: `(fileSize - headerSize) / sampleSize` and reading the last full sample's `t`.
- Binary download trims any trailing partial sample bytes and sets HTTP header `X-GaggiMate-Incomplete: 1`.
- Frontend parser (`parseBinaryShot.js`) infers count from payload length when header `sampleCount==0` and marks `incomplete: true`, using last sample `t` for effective duration. It also validates `sampleSize==48` and data alignment.

- WebSocket `req:history:list` now returns only metadata. 
- Eliminates repeated sprintf/println calls.
- Buffered 4 KB writes reduce flash wear and blocking time.
- Fixed-size header enables metadata listing without loading entire file.
- Constant per-sample layout simplifies JS parsing and avoids per-line parsing cost.
- Graceful recovery of interrupted shots avoids total data loss.
- Sample size stored in header allows strict validation (no silent corruption).
- Increase shot history from 3 to 10 entries
- Add storage use indicator in system & updates (in OTA response for simplicity)


src/display/plugins/ShotHistoryPlugin.cpp
```diff
@@ -18,52 +19,75 @@ void ShotHistoryPlugin::setup(Controller *c, PluginManager *pm) {
...
+                header.magic = SHOT_LOG_MAGIC;
+                header.version = SHOT_LOG_VERSION;
+                header.reserved0 = (uint8_t)SHOT_LOG_SAMPLE_SIZE; // record sample size actually used
+                header.headerSize = SHOT_LOG_HEADER_SIZE;
+                header.sampleInterval = SHOT_LOG_SAMPLE_INTERVAL_MS;
+                header.fieldsMask = SHOT_LOG_FIELDS_MASK_ALL;
+                header.startEpoch = getTime();
+                Profile profile = controller->getProfileManager()->getSelectedProfile();
+                strncpy(header.profileId, profile.id.c_str(), sizeof(header.profileId) - 1);
+                strncpy(header.profileName, profile.label.c_str(), sizeof(header.profileName) - 1);
```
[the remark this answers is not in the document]
jniebuhr: This seems like a valid note

src/display/plugins/WebUIPlugin.cpp
```diff
@@ -157,6 +164,46 @@ void WebUIPlugin::setupServer() {
     server.on("/api/scales/connect", [this](AsyncWebServerRequest *request) { handleBLEScaleConnect(request); });
     server.on("/api/scales/scan", [this](AsyncWebServerRequest *request) { handleBLEScaleScan(request); });
     server.on("/api/scales/info", [this](AsyncWebServerRequest *request) { handleBLEScaleInfo(request); });
+    server.on("/api/history", [this](AsyncWebServerRequest *request) {
```
jniebuhr: couldn't this whole method be replaced with serveStatic?

src/display/plugins/WebUIPlugin.cpp
```diff
@@ -21,6 +21,13 @@
 static std::unordered_map<uint32_t, std::string> rxBuffers;
 static WebUIPlugin* g_webUIPlugin = nullptr;
 
+// Safe JSON send helper: serialize to String to guarantee null termination
```
jniebuhr: I recently reworked it to use buffers since messages kept being cut off

src/display/plugins/WebUIPlugin.cpp
```diff
@@ -361,10 +405,8 @@ void WebUIPlugin::handleProfileRequest(uint32_t clientId, JsonDocument &request)
         }
     }
 
-    size_t bufferSize = measureJson(response);
-    auto *buffer = ws.makeBuffer(bufferSize);
-    serializeJson(response, buffer->get(), bufferSize);
-    ws.text(clientId, buffer);
+    // Allocate +1 for null terminator to avoid overwrite/corruption
+    wsSendJson(&ws, clientId, response);
```
jniebuhr: see above, this buffer was recently introduced

jniebuhr (changes requested):

PR #452, opened 2025-10-04
Fix script to safely check version

dragm83:
- Safely check version and set unknown if not found

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved a crash when version information is missing by safely checking for its presence.
  * Falls back to displaying "unknown" if no version is available, preventing errors.
  * Uses a computed local value for consistent version output.
  * Minor structural cleanup improves readability while maintaining behavior, enhancing robustness during analysis.


jniebuhr:
Already happened to me too and I fixed it

PR #484, opened 2025-11-08
Handle unicode when fetching shots from the machine

txsmith:
I noticed when trying to add an Allongé profile that there's some unicode weirdness. I could add the profile just fine, but when viewing it in the shot history it looks like this:

image

At first I thought it was primarily on the ESP32 side, because the é shows up as a `□` on the device. But on further inspection it seems that a simple Javascript fix at least makes unicode work on the web UI:

image

If you're interested I could take a look at fixing this on the display side as well, but i'm not super familiar with LVGL so would have to figure it out. Seems like the font used just doesn't have unicode in its range of characters.

## Summary by CodeRabbit

* **Refactor**
  * Improved UTF-8 string decoding in binary parsing operations for enhanced reliability and standards compliance in text handling.


jniebuhr:
I wasn't able to reach you for the CLA

PR #486, opened 2025-11-11
Sagemate purge valve

fj-sanchez:
## Summary by CodeRabbit

* **New Features**
  * Automatic purge valve control triggers on brew start/end with a timed purge.
  * PID tuning now includes an additional feed‑forward coefficient alongside Kp, Ki, Kd.

* **Changes**
  * BLE scale plugin set trimmed — most plugins disabled; BookooScales remains active.


jniebuhr:
Why are all the scales disabled? 
This cannot be merged like this as it changes existing behavior in a way that would not be wanted for a lot of machines, these options can be used as additional behavior

PR #503, opened 2025-11-28
Tare scale by holding touchui 'scales' button on brew/grind screens.

dragm83:
Tares the bluetooth scale and flow estimation by long-press/holding the touch UI scales button
image

## Release Notes

* **New Features**
  * Added long-press gesture support for volumetric buttons on brew and grind screens, enabling scale tare operations via press-and-hold interaction.
  * Long-press events now intelligently prevent accidental simultaneous click triggering, improving control responsiveness.


src/display/ui/default/lvgl/ui_events.cpp
```diff
@@ -145,3 +154,12 @@ void onProfileSaveAsNew(lv_event_t *e) {
     controller.onProfileSaveAsNew();
     controller.getUI()->changeBrewScreenMode(BrewScreenState::Brew);
 }
+
+void onVolumetricHold(lv_event_t * e)
+{
+    // Set flag to prevent click from firing when button is released
+    volumetricHoldTriggered = true;
```
jniebuhr, line 161: We don't have this on the brew button either, you're sure it's needed?

jniebuhr (changes requested):

src/display/ui/default/lvgl/ui_events.cpp
dragm83: When the brew button is held, the flush executes and the underlying button/screen changes so the 'onClick' function never executes since it executes on release.

PR #544, opened 2026-01-02
Countdown towards end of shot

aviv1ron1:
Currently the display during a shot counts the time passed since the shot began.
IMHO this is not an interesting nor helpful information.
I am interested to see a countdown how much time left towards the end of the shot.
I think this is much more helpful, and not a big change.

this pr also supports pro profiles, so for example if i have a pressure target and i finished a 10s step in 2s the countdown timer will reduce 10s from the time so that the countdown stays accurate.

If a shot takes more time then defined and the countdown reaches zero (can this even happen?) the counter will turn red and start counting upwards.

supported both on the display and the website.

## Release Notes

* **New Features**
  * Time-based brew processes now display an accurate countdown timer showing remaining brew time
  * Added automatic overtime detection with red color highlighting when brew duration exceeds the target time

* **Bug Fixes**
  * Improved reliability of brew process time calculations and status displays


aviv1ron1:
by the way, following the contributor.md I  ran the code formatters Iand  got like 50 files changed - files i did not touch. so I reverted these changes so the PR includes only my changes.

PR #553, opened 2026-01-18
Fix auto brew when switching from steam to menu

MajesticKoala:
Occasionally it is possible for the brew process to start when switching from an active steam screen (temp above target) to the menu.

See video replicating bug:
https://github.com/user-attachments/assets/5c52225e-fac5-428a-b8ec-e19e34cd6f7f

The issue has been identified as a race condition and can occur when Controller::loop() executes during the controller:mode:change event processing (<100ms window).

Moving `steamReady=false` to after the event, prevents loop() from activating in this code snippet:
```
if (mode == MODE_STEAM && !steamReady && currentTemp + 5.f > getTargetTemp()) {
    activate();
    steamReady = true;
}
```

## Summary by CodeRabbit

* **Refactor**
  * Internal code optimization for improved state management logic.


jniebuhr:
@cla-bot recheck

jniebuhr:
@cla-bot check

PR #569, opened 2026-02-02
Feature/gearpump

Velkromod:
With these changes, the objective is to improve the performance of the gear pump allowing to read the pulses from the pump ESC tacho pin. It counts time between pulses to infiere RPM (2 pulses per revolution, on falling from 5v to 80mv), It is coded to read pulses in the GPIO 13. From there on flowmap using the gear pump datasheet is inferred the actual flow using interpolation of the flow curves (intersection from RPM and pressure), then it converts ml/min to ml/sec, and feed this info into dimmed pump and pressure controller. With all these changes it should allow for a smooth pressure control tailored to the MPB 015 24V 70W Gear Pump. Since all this code was written by IA it's prone to have errors, so is to be treated as experimental.


jniebuhr:
Flow control / estimation can be achieved without RPM measurements, all these changes are not needed

PR #580, opened 2026-02-09
fix: Prevent WiFi from getting stuck offline after intermittent disconnects

jaapp:
Operator precedence bug caused the retry counter to never reset on successful reconnection, leading to premature AP mode activation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed WiFi connection state detection to properly recognize when the device transitions to a connected state, ensuring reliable WiFi connection handling.


jaapp:
Closing this PR as it is a duplicate of https://github.com/jniebuhr/gaggimate/pull/520 and also the wifimanager is never actually used in the codebase. So although this fix is real, it's of no consequence.

PR #599, opened 2026-02-17
Updated if statement so that if standby timeout is 0 it will never autmatically return to standby mode

dfligs:
Change to allow users to disable auto switch to standby mode after timeout by setting a value of 0 (disable)


jniebuhr:
@cla-bot check

PR #609, opened 2026-02-23
fix: Race condition, BLE buffer overread, spurious boot loop and controller waiting state

rartigues:
# Pull Request Draft

## Title
fix: Race condition, BLE buffer overread, spurious boot loop and controller waiting state

---

## Description

This PR addresses several stability issues found through coredump analysis and testing on a GaggiMate Pro Rev 1.1.

### Bug fixes

**Race condition on `currentProcess` between Core 0 and Core 1** (`Controller.cpp`)

`Controller::updateControl()` runs on Core 1 while `deactivate()` can set `currentProcess = nullptr` from Core 0. This caused a null pointer dereference that crashed the display firmware. Confirmed via ESP32-S3 coredump analysis — the backtrace showed `Controller::updateControl()` at the exact line reading from `currentProcess` without a guard.

Fix: capture `currentProcess` into a local variable at the start of every method that reads from it, so all accesses within that call see a consistent (possibly null) pointer.

Affected methods: `updateControl()`, `getTargetTemp()`, `isActive()`, `isGrindActive()`, `activate()`.

---

**BLE notify callback reads past buffer end** (`NimBLEClientController.cpp`)

`notifyCallback` was casting `pData` directly to `char*` and passing it to `atoi`, `atof`, `String()`, etc. BLE characteristic values are **not null-terminated**, so these calls could read past the end of the buffer. This is undefined behavior and was a source of intermittent crashes on received sensor data.

Fix: construct a bounded `std::string rawData((char*)pData, length)` once at the top of the callback and use that for all parsing.

---

**Null pointer dereference in `ControllerOTA::init()`** (`ControllerOTA.cpp`)

`ControllerOTA::init()` called `txChar->canNotify()` without first checking if `txChar` or `pRemoteService` were `nullptr`. If the OTA BLE service is not advertised by the controller, this would crash.

Fix: added null checks on `pRemoteService` and `txChar` before use.

---

**Spurious boot loop on controller when board is not detected** (`GaggiMateController.cpp`)

`detectBoard()` reads a voltage divider via ADC to identify the hardware revision. A single failed read (e.g., due to ADC noise at cold start, before the voltage regulator has fully settled) immediately triggers `ESP.restart()`, causing an intermittent boot loop that's hard to reproduce.

Fix: retry the ADC read up to 3 times with a 10 ms stabilization delay and a 500 ms pause between attempts before giving up and restarting.

---

### New feature

**Display shows "Waiting for controller..." when BLE controller is not reachable** (`Controller.cpp`, `Controller.h`, `DefaultUI.cpp`, `DefaultUI.h`, `NimBLEClientController.h`)

Previously, the display stayed on the init screen with "Starting..." indefinitely if the controller was not found during the initial BLE scan — sometimes triggering a boot loop. This gave no feedback to the user about the connection state.

Changes:
- After 10 seconds without a BLE connection, the init screen label changes to "Waiting for controller..."
- A new `controller:bluetooth:waiting` plugin event is fired, consistent with the existing event system
- The BLE scan is automatically restarted every ~10.5 seconds (one full scan duration) while the controller is not connected, so the display will pick up the controller as soon as it becomes available — no manual restart required
- When the controller eventually connects, the label resets and the screen transitions normally
- `BLE_SCAN_DURATION_SECONDS` promoted to a `public static constexpr` on `NimBLEClientController` so the retry interval can be derived from it

---

## Testing

Tested on GaggiMate Pro Rev 1.1 hardware with display (LilyGo T-RGB) and controller connected simultaneously via USB.

- **Race condition**: Rapid activate/deactivate cycling via the touchscreen did not produce any crashes over extended testing. Previously this would crash within minutes.
- **BLE buffer**: No crashes observed on sensor data reception. Previously triggered by certain controller firmware notification payloads.
- **Board detection**: Power-cycled the controller ~20 times in succession; board was detected correctly on every boot.
- **Waiting state**: Powered the display without the controller connected → "Starting..." → "Waiting for controller..." after ~10s. Connected controller to power → display detected it within the next scan cycle and transitioned to the brew screen without any restart.

---

## Commits

```
fix: race condition on currentProcess between Core 0 and Core 1
fix: use length-bounded string construction in BLE notify callback to prevent buffer overread
fix: retry board detection up to 3 times with ADC stabilization delay to prevent spurious boot loop
feat: show 'Waiting for controller...' on init screen when BLE controller is not found within timeout
fix: periodically restart BLE scan while waiting for controller to appear
```

## Release Notes

* **New Features**
  * Enhanced Bluetooth connection feedback with "Waiting for controller..." status display during pairing.
  * Automatic BLE scan management with intelligent timeout and restart handling.

* **Bug Fixes**
  * Improved board detection reliability with automatic retry mechanism.
  * Added null safety checks for OTA service initialization to prevent crashes.

### DIFF
--- a/.clabot
+++ b/.clabot
@@ -1,4 +1,4 @@
 {
-  "contributors": ["jniebuhr", "dragm83", "DevNullGamer", "nayrod485", "spsancti", "Sartoric", "nodu", "modsmthng", "MajesticKoala", "njtl", "Serokz", "jaapp", "dfligs"],
+  "contributors": ["jniebuhr", "dragm83", "DevNullGamer", "nayrod485", "spsancti", "Sartoric", "nodu", "modsmthng", "MajesticKoala", "njtl", "Serokz", "jaapp", "dfligs", "rartigues"],
   "message": "We require contributors to sign our Contributor License Agreement, and we don't have yours on file. In order for us to review and merge your code, please contact @jniebuhr (mdwasp) on Discord to get yourself added."
 }


jniebuhr:
@cla-bot check

PR #611, opened 2026-02-24
Fix BLE connection race condition and add visual status indicator

jniebuhr:
## New Features
- Added Bluetooth disconnect event handling with automatic reconnection retry mechanism
- Enhanced UI now displays accurate Bluetooth connection status on initialization and standby screens
- Improved state management for connection and disconnection events

## Bug Fixes
- Added safety guards to prevent invalid state handling and crashes during status updates


jniebuhr:
@cla-bot check

PR #622, opened 2026-03-07
fix: fix being unable to type decimals in extended phase targets

kraklo:
Fixes being unable to type in decimals in extended phase targets.

The issue was due to the input being a `number`, `e.target.value` is implicitly passed through `parseFloat`. What this would result in is:

- User types in an incomplete decimal (e.g. `3.`)
- `onChange` fires
- `e.target.value` = `3`
- `onChange` prop callback fires, but because a new object ref is created, that changes `phase.targets` a few layers up
- Due to props changing, component re-renders, passing in `target.value` (which is still 3), effectively eating the decimal input

The fix here is to only do the callback if the parsed value actually changes. This means that a decimal can be typed now. It's still a little janky when deleting (i.e. going from `3.1` -> `3.`) as it gets parsed and deletes the decimal point.

## Summary by CodeRabbit

* **Bug Fixes**
  * Numeric input in the Profile Editor now avoids redundant change events: updates only trigger when the value actually changes, reducing unnecessary processing.
  * Typing or pasting an identical value no longer invokes duplicate callbacks, improving responsiveness and preventing repeated save/validation actions.

### DIFF
--- a/web/src/pages/ProfileEdit/ExtendedPhaseTarget.jsx
+++ b/web/src/pages/ProfileEdit/ExtendedPhaseTarget.jsx
@@ -59,12 +59,16 @@ export function ExtendedPhaseTarget({ onChange, target, index, onRemove }) {
                   className='grow'
                   type='number'
                   value={target.value || 0}
-                  onChange={e =>
-                    onChange({
-                      ...target,
-                      value: parseFloat(e.target.value),
-                    })
-                  }
+                  onChange={e => {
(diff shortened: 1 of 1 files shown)


hutchinsp01:
+1
This is a very minor annoyance, but having to copy and paste decimals in sucks

hutchinsp01:
@jniebuhr Can you bring this back?

PR #624, opened 2026-03-10
Add pressure offset for machines with grouphead mushroom valve

stephenwilley:
Some machines use a mushroom valve instead of a 3-way solenoid to open the grouphead (for example, the Delonghi ECP).  This will open at a particular pressure (4 bar in the case of the ECP) and effectively reduce the pump pressure onto the puck by that value.

This PR allows an offset to be used during brew mode that compensates for that, meaning you don't have to modify profiles to suit these machines.  It clamps the pressure values to 0.0 to prevent the UI from presenting a negative pressure.

I've tried to stick closely to the methodology used for temperatureOffset.  UI change screengrab attached.  I'll also attach a pressure test utility profile I created to determine the cracking pressure for the valve.

(I only have an ECP to test this on.  Given the defaults and the code it shouldn't change any behaviours without the user explicity setting the offset value, but of course I would appreciate any testing anyone might want to do on other machines)

pressure_offset
[Utility-Pressure-Test.json](https://github.com/user-attachments/files/25884718/Utility-Pressure-Test.json)

## Summary by CodeRabbit

* **New Features**
  * Pressure Offset setting added and now applied to live pressure readings during brew.
  * Developer Mode toggle added; reveals manual update UI and exposes developer flag in status.
  * Local OTA upload from the web UI for firmware/filesystem/controller with improved progress reporting and restart handling.

* **UI**
  * Standby time display hides leading zero in 12‑hour mode.

### DIFF
--- a/src/display/core/Controller.cpp
+++ b/src/display/core/Controller.cpp
@@ -227,9 +227,6 @@ void Controller::setupWifi() {
             configTzTime(resolve_timezone(settings.getTimezone()), NTP_SERVER);
             setenv("TZ", resolve_timezone(settings.getTimezone()), 1);
             tzset();
-            sntp_set_sync_mode(SNTP_SYNC_MODE_SMOOTH);
-            sntp_setservername(0, NTP_SERVER);
-            sntp_init();
         } else {
             WiFi.disconnect(true, true);
             ESP_LOGI(LOG_TAG, "Timed out while connecting to WiFi");


stephenwilley:
Oh, very sorry.  I've been pushing more commit than I intended to this.  I'll make a new PR and redo it properly.  My bad.

PR #625, opened 2026-03-12
Add pressure offset using temperature offset as a template

stephenwilley:
Some machines use a mushroom valve instead of a 3-way solenoid to open the grouphead (for example, the Delonghi ECP). This will open at a particular pressure (4 bar in the case of the ECP) and effectively reduce the pump pressure onto the puck by that value.

This PR allows an offset to be used during brew mode that compensates for that, meaning you don't have to modify profiles to suit these machines. It clamps the pressure values to 0.0 to prevent the UI from presenting a negative pressure.

I've tried to stick closely to the methodology used for temperatureOffset. UI change screengrab attached. I'll also attach a pressure test utility profile I created to determine the cracking pressure for the valve.

(I only have an ECP to test this on. Given the defaults and the code it shouldn't change any behaviours without the user explicity setting the offset value, but of course I would appreciate any testing anyone might want to do on other machines)

Settings UI Screenshot:
pressure_offset

Pressure Test Profile:
[Utility-Pressure-Test.json](https://github.com/user-attachments/files/25924184/Utility-Pressure-Test.json)

## Summary by CodeRabbit

* **New Features**
  * Added a configurable Pressure Offset setting (saved via web UI) to fine‑tune pressure measurements.
  * Settings UI now shows a "Pressure Offset (bar)" field when the machine reports pressure capability.

* **Behavior Changes**
  * Pressure readings and brew control now apply the configured offset so displayed and controlled pressures reflect user adjustment.


scorpionleonid (approved):

jniebuhr:
@cla-bot check

vasto7:
Nice feature — this would be really helpful for Delonghi ECP and similar mushroom-valve machines.

Two questions:

1. Does the offset also shift the pressure seen by the pump flow model (`getAvailableFlow()` evaluates at `_filteredPressureSensor`)? If so, users would need to recalibrate their pump flow coefficients after enabling the offset, right?
2. Are there any potential issues with the auto calibration wizard when the offset is active?

PR #628, opened 2026-03-17
feat: Add connection check for weighing scale before brewing

0xStuart:
If a paired weighing scale is missing, or connected but not yet sending weight, brewing is blocked and a dialog offers **Ignore** or **Cancel**. Waiting is implicit: the shot starts once a live weight sample arrives.

Scale wait dialog

Rewritten on current `master` after the EEZ UI migration. The dialog is a transient LVGL overlay on `lv_layer_top()` rather than an EEZ screen, so it can sit on top of brew without a Studio page and so EEZ rebuilds do not clobber it.

GATT connect alone is not enough: until the first BLE weight sample, `isBluetoothScaleHealthy()` is false and the brew would start as a timed shot that then ignores later samples.

Tested on device.

### DIFF
--- a/src/display/core/Controller.cpp
+++ b/src/display/core/Controller.cpp
@@ -977,9 +977,31 @@ void Controller::updateControl() {
     controlStateSent = true;
 }
 
-void Controller::activate() {
+void Controller::activate(bool ignoreScale) {
     if (isActive())
         return;
+
+#ifndef GAGGIMATE_SIM
+    // A paired scale or a volumetric profile needs a live scale. GATT connect is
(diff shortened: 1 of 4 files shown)


0xStuart:
Rewrote this against current master after the EEZ migration. Same feature, but:

- The dialog is now a transient LVGL overlay (not an EEZ / SquareLine screen) so it can sit on top of brew.
- Auto-start waits for a real weight sample, not just BLE GATT connect — otherwise the shot starts timed and later BLE samples are ignored.

Happy to redo the overlay in EEZ Studio if that is preferred. CLA is still outstanding from March if that needs signing.

PR #629, opened 2026-03-18
fix: semver check for nightly releases 

ggazzo:
I noticed that after updating to the latest nightly, it sometimes shows as if there is another update available. Looking at the code, I believe this could be a potential fix.

I believe the issue is that the prerelease comparison is done character by character. If we don’t properly copy the null terminator (\0), any garbage in memory may lead to incorrect comparison results.

This kind of issue is somewhat tricky to reproduce in real life scenarios, as it would require building and testing directly in my repository and change all pointing to there. However, I did run some unit tests to validate the behavior. I chose not to include the full test suite changes in this PR to keep small as possible.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory safety in OTA version parsing: safer prerelease handling to avoid malformed reads, only processes prerelease data when present, and ensures a safe early return on allocation failure to reduce risk of update errors or crashes.

### DIFF
--- a/lib/OTA/src/semver_extensions.cpp
+++ b/lib/OTA/src/semver_extensions.cpp
@@ -37,6 +37,9 @@ semver_t from_string(const string &version) {
         size_t n = prerelease.length();
         prerelease_ptr = (char *)calloc(n + 1, 1);
         if (n > 0) {
+            if (prerelease_ptr == nullptr) {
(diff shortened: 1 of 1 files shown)


ggazzo:
Resolved by #678 — the same prerelease buffer overflow was fixed there (malloc N+1 + explicit null terminator + null check). Closing.

PR #652, opened 2026-04-03
Remove redundant sntp code and turn off smooth corrections

stephenwilley:
The existing setup uses SNTP_SYNC_MODE_SMOOTH which only corrects the time really really slowly unless you're more than 35 minutes out.  On boot of a coffee machine it should probably just instantaneously correct.

The other two lines are just redundant since they're already taken care of by configTzTime

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified Wi‑Fi startup: timezone configuration on successful Wi‑Fi remains, while automatic network time client initialization during startup has been removed to streamline connection flow and avoid redundant steps. Connection, fallback, and update behaviors remain unchanged.

### DIFF
--- a/src/display/core/Controller.cpp
+++ b/src/display/core/Controller.cpp
@@ -235,9 +235,6 @@ void Controller::setupWifi() {
             configTzTime(resolve_timezone(settings.getTimezone()), NTP_SERVER);
             setenv("TZ", resolve_timezone(settings.getTimezone()), 1);
             tzset();
-            sntp_set_sync_mode(SNTP_SYNC_MODE_SMOOTH);
-            sntp_setservername(0, NTP_SERVER);
-            sntp_init();
         } else {
             WiFi.disconnect(true, true);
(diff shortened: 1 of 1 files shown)


jniebuhr:
@cla-bot check

jniebuhr:
Wouldn't the coffee machine be more than 35 minutes out on the first boot? time set to 1970

stephenwilley:
configTzTime already calls sntp_init so it won't 'stay wrong'
https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-time.c

The smooth time thing is designed so that really time-sensitive stuff corrects without any big jumps if you're within the 35 minute limit - useful for payment processors etc. but for a coffee machine, people just want the correct time as fast as possible.

If you're outside the 35 minute difference then yeah it moves quickly, but as I noticed, my clock was only wrong by about 10 minutes at one point (while I was cautiously testing my machine, I would pull the mains plug frequently and I think that would make the clock be 'a bit wrong' when it came back on).

PR #664, opened 2026-04-17
feat: Add configurable flush duration setting

marxd262:
- Add flushDuration preference (1-60s, default 5s) persisted to NVS
- Add getter/setter methods in Settings class
- Expose setting via WebUI
- Apply duration dynamically to FLUSH_PROFILE instead of using hardcoded value
- Default duration is still 5s

# Reason
When using a slower Pump (Olab Silent Green) the amount of water when flushing into a cup to heat it up can be less then ideal. So multiple flushes may be needed. This can be frustrating when using bigger cups. This feature allows the user to increase the amount of time that the Flush profile runs. This only changes the max amount. When releasing the play button on the display this will still stop the flush.

# Changes in WebUI
The UserPreferences card now includes a field for Flush Duration. 
image

## Summary by CodeRabbit

* **New Features**
  * Added "Flush Duration" setting in User Preferences (numeric seconds, 1–60, default 5s).
  * Setting is persisted and restored across sessions.
  * Web UI exposes and accepts the new value.
  * Flush operations now use the configured duration when started.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jniebuhr/gaggimate/pull/664?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


marxd262:
I checked the complained of SonarCloud: [Do not assign data members in a constructor. Initialize member "flushDuration" in an initialization list.](https://sonarcloud.io/project/issues?pullRequest=664&open=AZ2bHs7NlbRe63qrhroj&id=jniebuhr_gaggimate)

As far as I can see, the implemented approach seems to be consistent with the other preferences that get saved and loaded. Implementing another approach would reduce maintainability as there would then be multiple ways to maintain. Implementing another way for all parameters that would follow sonar clouds suggestion seems to be outside of the scope of this PR. 

Please correct me if I overlooked something.

jniebuhr:
Same feature done in different PR

PR #666, opened 2026-04-17
schema: rewrite profile.json as self-documenting prescriptive contract

thetimetraveler:
## Summary

Rewrites `schema/profile.json` to mirror the actual behavior of the C++ parser in `src/display/models/profile.h` and serve as authoritative human-readable documentation for the profile JSON format.

The current schema has several fields where it disagrees with the parser, and it lacks descriptions on almost every property. This rewrite fixes the divergences and adds comprehensive inline documentation (with file:line citations into the firmware source) so a profile author can work directly from the schema without reading C++.

## Fixes to parser-vs-schema divergence

- `transition.adaptive` — was `integer [0, 1]`; parser reads `.as<bool>()` at `profile.h:265`, and all reference profiles (`data/p/*.json`) write JSON booleans. Now `"type": "boolean"`.
- Simple `pump` — was `enum [0, 1]`; web UI accepts `min=0 max=100` (`ExtendedPhase.jsx:273-274`), built-in flush profile uses `100`. Now `"type": "integer", "minimum": 0, "maximum": 100`.
- `utility` — parsed by firmware (`profile.h:222`) and written by `writeProfile` (`profile.h:314`), but missing from the schema. Caused `additionalProperties: false` to reject `data/p/flush.json`. Now declared as optional boolean.
- `id` — was `required` with `"format": "uuid"`; parser tolerates missing (`profile.h:214-215`), `ProfileManager.cpp:114-117` auto-generates via `generateShortID()` which produces a 10-char `[0-9A-Za-z]` string, not a UUID. Now optional with pattern `^[A-Za-z0-9_-]+$`.
- `target.type` — the allowed values `volumetric | pressure | flow | pumped` were already correct, but the schema gave users no warning that unknown strings (notably `"weight"`, which the shot-analyzer web UI renders as a display alias) are silently dropped by `continue;` at `profile.h:287-289`. Now documented explicitly.

## Documentation additions

- `$id` with a versioned URI so downstream validators / SchemaStore consumers can pin.
- `description` on every field with units stated inline (bar, g/s, ml/s, g, ml, s, °C) and file:line citations into the parser or BrewProcess source for every non-obvious claim.
- Every enum uses `oneOf` + `const` + `title` + `description` branches (rendered by json-schema-for-humans, Docson, IntelliJ) with a parallel `enumDescriptions` array (rendered in VS Code hovers). This is the portable cross-tool pattern documented in [JSON Schema spec issue #1062](https://github.com/json-schema-org/json-schema-spec/issues/1062) for per-enum descriptions.
- Sentinel values explicitly documented: pump `pressure/flow: -1` = maintain measured value at phase start; pump limit `: 0` = ignore this limit; `phase.temperature: 0` = fall back to profile temperature; volumetric `target.value: 0` = no stop (run full duration).
- Runtime divergence between `standard` and `pro` profiles (volumetric-target blocks duration timeout only for standard, per `profile.h:109-111`) is called out on both the `type` enum and the phase `duration` field.
- `BREW_SAFETY_DURATION_MS = 300s` hard cap from `constants.h:10-11` is noted on the phase `duration` maximum.

## Structure

- Extracted `phase`, `transition`, `pump` (the oneOf union), `pumpSimple`, `pumpAdvanced`, and `target` into `definitions/` for reuse and top-level readability.
- Added `patternProperties: { "^_": {} }` as an extension escape so users can annotate profiles with `_notes`, `_source`, `_bean`, etc. without breaking `additionalProperties: false` strict validation. Forward-compat pattern borrowed from compose-spec (`"^x-"`).

## Validation

All five bundled reference profiles pass the new schema under ajv:

```
data/p/9bar.json     valid
data/p/adapt.json    valid
data/p/flush.json    valid
data/p/lever.json    valid
data/p/lmleva.json   valid
```

Negative tests to confirm the schema is actually strict:

- Profile with `"type": "weight"` target → rejected with _"must be equal to one of the allowed values"_.
- Profile with `"adaptive": 1` (integer) → rejected with _"must be boolean"_.

No firmware / parser / UI code is changed by this PR — it is schema-only.

## Test plan

All checks below were executed locally against this branch at 1def98f.

- [x] **`npx ajv-cli@5 validate -s schema/profile.json -d "data/p/*.json" --strict=false` returns valid for all five reference profiles.** Result: all five (`9bar`, `adapt`, `flush`, `lever`, `lmleva`) report `valid`.
- [x] **Schema compiles as a valid JSON Schema** (prerequisite for any IDE to load it for hover tooltips). `npx ajv-cli@5 compile -s schema/profile.json --strict=false` returns `schema schema/profile.json is valid`. Reviewers can confirm the editor-hover experience by opening any `data/p/*.json` in VS Code with `"$schema": "./schema/profile.json"` added to the file — every field and enum value surfaces its description on hover.
- [x] **Typo detection verified.** Copied `data/p/9bar.json` to a scratch path, changed `targets[0].type` from `"volumetric"` to `"weight"`, and ran ajv. Result: `invalid — data/phases/0/targets/0/type must be equal to one of the allowed values`. This is the specific failure mode the prior schema silently accepted; the new schema catches it before it reaches the firmware (where `profile.h:287-289` would silently drop the target via `continue;`).
- [x] **HTML documentation renders.** `generate-schema-doc schema/profile.json /tmp/profile.html` (from [json-schema-for-humans](https://github.com/coveooss/json-schema-for-humans)) produced a 167 KB browsable page. Title, enum branch descriptions, `definitions` cross-references, file:line citations into `profile.h`, and sentinel value descriptions all render correctly. The HTML is a generated artifact and is **not** committed — it should be regenerated in CI and optionally published to GitHub Pages (see follow-ups).

## Out of scope / possible follow-ups

- Add `.vscode/settings.json` pointing `json.schemas` at `./schema/profile.json` for auto-validation in editor.
- Add a `check-jsonschema` pre-commit hook running against `data/p/*.json`.
- Add a docs build step (e.g., CI job running `generate-schema-doc`) that publishes the HTML reference — either to GitHub Pages or as a downloadable artifact on releases. Committing the HTML directly is **not recommended** since it is regenerated from the schema on every change.

## Summary by CodeRabbit

* **New Features**
  * Profile `id` is now optional
  * Added `utility` flag for profiles (default: false)
  * Reserved annotation keys (starting with _) are supported
  * Improved pump options: simple percentage or advanced object form

* **Changes**
  * Stricter validation and sensible defaults (temperature bounds, min label length, favorites/selection defaults)
  * Phases now require at least one item and stronger phase, transition, and target constraints (operator, non‑negative values, bounded durations)

### DIFF
--- a/schema/profile.json
+++ b/schema/profile.json
@@ -111,11 +111,22 @@
           "description": "Phase duration (hard cap) in **seconds**. The phase exits when duration elapses if no target fires first. Firmware caps each phase at 300 s via `BREW_SAFETY_DURATION_MS` (`src/display/core/constants.h:10–11`). Minimum enforced by the helper `adjustDuration` is `0.5s` (`src/display/models/profile.h:71`); the web UI enforces `min=1`."
         },
         "temperature": {
-          "type": "number",
-          "minimum": 0,
-          "maximum": 160,
           "default": 0,
-          "description": "Per-phase boiler setpoint in °C. **Sentinel: `0` means use the profile-level `temperature`** (`src/display/core/process/BrewProcess.h:128–133`). Any non-zero value overrides the profile default for the duration of this phase. Firmware safety range is `MIN_TEMP=0` to `MAX_TEMP=160`."
+          "description": "Per-phase boiler setpoint in °C. Either the `0` sentinel (inherit the profile's top-level `temperature`) or an explicit positive value up to the firmware cap.",
+          "oneOf": [
+            {
+              "const": 0,
+              "title": "Use profile default",
+              "description": "Sentinel: inherit the profile-level `temperature` field for the duration of this phase (`src/display/core/process/BrewProcess.h:128-133`). This is the default for new phases."
+            },
+            {
+              "type": "number",
+              "exclusiveMinimum": 0,
+              "maximum": 160,
+              "title": "Override temperature in °C",
+              "description": "Explicit phase temperature in °C. Firmware safety cap MAX_TEMP=160 (`src/display/core/constants.h:16`). Typical espresso range is 85–96 °C."
+            }
+          ]
         },
         "transition": {
           "$ref": "#/definitions/transition",
--- a/schema/profile.json
+++ b/schema/profile.json
@@ -1,187 +1,261 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "title": "Espresso Profile Schema",
+  "$id": "https://github.com/jniebuhr/gaggimate/schema/profile/v1.json",
+  "title": "GaggiMate Espresso Profile",
+  "description": "Profile definition consumed by the GaggiMate firmware. A profile is an ordered sequence of phases that the machine executes end-to-end. Each phase runs until one of its stop conditions fires (duration timeout, or any matched target). This schema is authored from the C++ parser at `src/display/models/profile.h` and is intended to be used as both the validation contract and the canonical documentation for the file format. When the schema and the parser disagree, the parser wins and the schema is a bug.",
   "type": "object",
   "additionalProperties": false,
+  "patternProperties": {
+    "^_": {
+      "$comment": "Reserved for user annotations (e.g. _notes, _source, _bean). Ignored by the firmware parser; preserved in on-disk JSON but not round-tripped through the save path."
+    }
+  },
+  "required": ["label", "type", "description", "phases"],
   "properties": {
     "id": {
       "type": "string",
-      "format": "uuid"
+      "pattern": "^[A-Za-z0-9_-]+$",
+      "description": "Stable identifier used as the on-disk filename (`<id>.json`) and referenced from the shot history. Optional on import: if omitted, the firmware generates a 10-character alphanumeric ID via `generateShortID()` (`src/display/core/utils.cpp:9`). Must be filesystem-safe — only letters, digits, underscore, and hyphen.",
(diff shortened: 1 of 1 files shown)


thetimetraveler:
Thanks @mdwasp — fixed in 1fb07c9.

Updated the `id` description to state explicitly: _"Do not set this in hand-authored or imported profiles. The web UI deletes this field on new imports, exports, and duplicates."_ Also added `"readOnly": true` as a portable hint so editors dim the field and form-generators disable the input.

The `pattern` constraint is kept (rather than removing the field) so firmware-exported profiles still round-trip cleanly — the schema documents what shape the firmware produces when it does write an id.

thetimetraveler:
Addressed in efc90c6.

`pumpAdvanced.pressure` and `pumpAdvanced.flow` now use `oneOf`:

- `{ "const": -1, "title": "Maintain current...", "description": "..." }` — the sentinel
- `{ "type": "number", "minimum": 0, "maximum": 12|15, "title": "Explicit ... in bar|g/s", "description": "..." }` — the real range

Each branch gets its own title and description, so json-schema-for-humans / Docson / IntelliJ render two distinct options with their semantic meaning rather than one combined "-1 to 15" range.

Boundary checks:
- `-1` → valid (sentinel)
- `0`, `7`, `12` → valid (in range)
- `-0.5`, `-2` → **rejected** (not the sentinel, not in range)
- `13` → rejected (over max)

All five `data/p/*.json` reference profiles still validate.

Issue #61, opened 2025-01-25
Reconnecting to display after restarting


peteger:
When I keep the display powered (e.g. via USB-C) and turn off/on just the PCB with the machine, connection is dropped and I need to restart also display.

jniebuhr:
Fixed in #174

Issue #340, opened 2025-07-31
[Feature Request] Temperature Profiling


georgigriffiths:
### Describe the feature

Would like to set a target temperature for each stage of the profile

### Why is this feature useful?

https://www.ranciliogroup.com/temperature-profiling-the-magic-touch-for-the-perfect-espresso/

Extract from the link
If we wish to enhance the acid notes of a delicate, high altitude, medium roast coffee, we should use an ascending profile. This curve would enable us to properly extract the lipid part as well and enhance density and body. On the other hand, if we wish to bring out the roasted notes of a natural coffee or a medium roasted robusta, avoiding any unpleasant bitterness and astringency, we should use a decreasing temperature curve. In this way we would obtain a more pronounced extraction of sweetness or acidity in the initial phase, while decreasing bitterness and above all astringency.

### How complex is this feature?

??? - I have absolutely no idea - so most likely VERY HIGH!

jniebuhr:
Duplicate of #240

PR #671, opened 2026-04-18
fix(bookoo): parse full weight notification, use recommended commands, guard ounce mode

thetimetraveler:
Was debugging an unrelated brew-by-weight issue and dug into the Bookoo protocol spec ([mini](https://github.com/BooKooCode/OpenSource/blob/main/bookoo_mini_scale/protocols.md), [ultra](https://github.com/BooKooCode/OpenSource/blob/main/bookoo_ultra_scale/protocols.md)) — realized the driver was throwing away ~80% of each 20-byte weight notification. This PR parses the rest and wires the useful fields through to the firmware.

> **Status (2026-04-19):** driver changes now upstream as [gaggimate/esp-arduino-ble-scales#29](https://github.com/gaggimate/esp-arduino-ble-scales/pull/29). This firmware PR will be rebased to consume the upstream library once that merges; the vendored copy here is transitional. Bug fixes from [@salekseev](https://github.com/salekseev)'s review are mirrored into both PRs. See the updated commit list below.

## What the Bookoo packet actually contains

Bytes 0-1 are the `0x03 0x0B` header. Bytes 2-4 are a scale-internal timer (ms). Byte 5 is weight unit (`0x01`=oz, `0x02`=g). Bytes 6-9 are the signed weight × 100. **Bytes 10-12 are signed flow rate × 100 (g/s) — scale computes it natively.** Byte 13 is battery %. 14-15 standby timer, 16 buzzer, 17 flow-smoothing-enabled. Byte 18 is "auto-mode stop condition" on Ultra, reserved on Mini. 19 is the XOR checksum.

Before this PR: only weight sign + weight were used. Everything else on the floor.

## Commands changed

- Tare: switched from `0x01` (plain tare) to **`0x07` (Tare + Start Timer)** — Bookoo's recommended command. It atomically resets weight AND marks the scale as "shot in progress", which prevents auto-sleep mid-brew. Firmware semantics unchanged (we don't consume scale timer yet).
- New `disableScaleSmoothing()` sends `0x08 0x00` on connect. The scale runs its own EMA on flow; `ShotHistoryPlugin` runs another EMA (α=0.25) on top of that. Double-smoothing is strictly worse than either alone — turning off the scale-side filter gives the firmware access to raw per-sample flow.

## Firmware plumbing

`RemoteScales` grows five optional accessors (`getFlowRate`, `getBatteryLevel`, `getScaleTimerMs`, `getWeightUnit`, `getAutoModeStopCondition`), each with a `hasX()` capability flag defaulting to false. Non-Bookoo drivers (Acaia, Decent, Felicita, Timemore, Varia, WeighMyBrew, MyScale, Difluid, Eclair, Eureka) compile unchanged — they just don't opt in.

`BLEScalePlugin` proxies the getters, emits `controller:volumetric-measurement:scale-flow:change` alongside each weight measurement (~10 Hz), and polls battery/unit at 1 Hz, firing `scale:battery:change` / `scale:weight-unit:change` only on value transitions.

`Controller::isVolumetricAvailable()` now returns false if the scale reports ounces. Raw values in oz mode are ~28× off — the volumetric stop would fire either immediately or never, and the shot would be silently wrong. `UNKNOWN` (driver doesn't parse unit) is treated as grams for backwards compat. `BLEScalePlugin::onMeasurement` also checks the unit per sample and aborts volumetric mid-shot if the user physically flips the scale to oz while brewing, falling through to time-based stop.

I deliberately left `hasAutoModeStopCondition()` returning false because byte 18 is reserved=0x00 on Mini but meaningful on Ultra, and there's no clean discovery path to tell them apart. Parsed into the accessor anyway so an Ultra-specific subclass could surface it later.

## Why the library is vendored (transitional)

The vendored copy under `lib/esp-arduino-ble-scales/` lets the driver changes be reviewed in the same diff as the firmware consumers that need them. The driver changes are now open upstream as [gaggimate/esp-arduino-ble-scales#29](https://github.com/gaggimate/esp-arduino-ble-scales/pull/29); once that merges, the vendored tree will be deleted and `platformio.ini` restored to use the upstream library. See `lib/esp-arduino-ble-scales/VENDORING.md` for details.

## Commits

**Original (landing the feature):**
- `d383301b` vendor the library
- `7d40345b` extend `RemoteScales` abstraction
- `59d84705` Bookoo driver: parse full notification + `0x07` tare + disable scale smoothing
- `fa4898c0` `BLEScalePlugin` exposes native fields and broadcasts events
- `3308c011` `Controller` refuses volumetric when scale is in ounce mode

**Review fixes ([@salekseev](https://github.com/salekseev) on 2026-04-19):**
- `56181186` `fix(bookoo): frame per call + erase processed + flip decode loop sentinel` (bug 1 + risk 4)
- `903eb31a` `fix(bookoo): ignore inbound SYSTEM instead of re-taring` (bug 2)
- `521f3d57` `refactor(bookoo): drop dead msgType param from sendMessage` (risk 6)
- `1d1413cb` `fix(remote_scales): virtual destructor for polymorphic unique_ptr` (bug 3)
- `7361a5c6` `fix(ble-scale): mid-shot ounce-mode abort via per-sample guard` (risk 5)
- `41b99e84` `docs(vendoring): reference upstream PR + correct stale commit hash`

**Scale-timer sync (follow-up feature from the same thread):**
- `f900e871` `feat(scales): add timer-control abstraction on RemoteScales base`
- `33fcae6f` `feat(scales/bookoo): implement startTimer/stopTimer/resetTimer`
- `dbab99c5` `feat(ble-scale): stop scale timer on brew end`

The timer wiring uses the existing `tare()` (0x07 = tare + start timer) at brew-prestart to start the scale's stopwatch, then calls `stopTimer()` (0x05) on `controller:brew:end` to freeze the final shot time on the scale's display. On Bookoo Ultra this requires the scale to be physically set to timing-mode (no BLE mode-switch command exists); Mini has no such restriction. Other drivers' `hasTimerControl()` defaults to false so they're unaffected.

## Hardware verification

Flashed on Gaggia Classic + GaggiMate Pro + Bookoo Themis Ultra. Shot #61 pulled **35.8g on a 36g target** (Decaf v5 profile, 1:2 ratio, 51.9s total). Volumetric routing triggers correctly, 5-phase profile runs clean, all the new `hasX()` capability flags report true, scale flow/battery/unit events emit as expected.

One gotcha worth flagging for anyone else touching the vendored library: `pio run -e display` is incremental and only rebuilds the files that changed. When the header's struct layout changes (new fields in `RemoteScales`), the library's `remote_scales.cpp.o` isn't rebuilt and still has the old layout — writes to the new fields then clobber unrelated memory, and BLE weight notifications silently stop reaching the display. **Always `pio run -e display -t clean` once after editing `remote_scales.h`** or you'll chase a ghost.

## Out of scope that I noted while writing this

- Shot log binary format bump for `svf` (scale-native flow) — touches the web UI decoder, better as its own PR.
- Web UI consumers for `scale:battery:change` etc. — events exist, hookup is future work.
- `UltraBookooScales` subclass to surface auto-mode stop condition safely.
- Three nits from the external review deferred to a follow-up PR: flow/unit events emitting at 10 Hz regardless of change, metadata getters polled at 10 Hz instead of 1 Hz, `disableScaleSmoothing` not resent after reconnect.

### DIFF
--- a/lib/esp-arduino-ble-scales/src/remote_scales.h
+++ b/lib/esp-arduino-ble-scales/src/remote_scales.h
@@ -70,7 +70,7 @@ public:
   virtual void disconnect() = 0;
   virtual void update() = 0;
 
-  ~RemoteScales() { clientCleanup(); }
+  virtual ~RemoteScales() { clientCleanup(); }
 protected:
   RemoteScales(const DiscoveredDevice& device);
   const DiscoveredDevice& getDevice() const { return device; }
--- a/lib/esp-arduino-ble-scales/src/remote_scales.h
+++ b/lib/esp-arduino-ble-scales/src/remote_scales.h
@@ -21,12 +21,41 @@ private:
   int rssi;
 };
 
+// Weight unit reported by the scale. Some BLE espresso scales can switch to
+// ounces, which would silently corrupt volumetric targets unless the firmware
+// detects the unit. UNKNOWN is the conservative default for drivers that don't
+// report the field.
+enum class ScaleWeightUnit : uint8_t { UNKNOWN = 0, GRAM = 1, OUNCE = 2 };
+
+// Sentinel for "driver has no battery reading available".
+constexpr uint8_t REMOTE_SCALES_BATTERY_UNKNOWN = 0xFF;
+
 class RemoteScales {
 
 public:
   using LogCallback = void (*)(std::string);
 
+  // Core weight (always available).
   float getWeight() const { return weight; }
+
+  // Optional native fields. Drivers that parse these should call the matching
+  // protected setters from their notification handler AND override the
+  // matching hasX() virtuals to return true. Defaults are safe no-ops for the
+  // drivers (Acaia, Decent, Felicita, Timemore, ...) that only parse weight.
+  float getFlowRate() const { return flowRate; }            // g/s, native from scale when available
+  uint8_t getBatteryLevel() const { return batteryLevel; }  // 0-100 %, or REMOTE_SCALES_BATTERY_UNKNOWN
+  uint32_t getScaleTimerMs() const { return scaleTimerMs; } // scale's internal stopwatch
+  ScaleWeightUnit getWeightUnit() const { return weightUnit; }
+  uint8_t getAutoModeStopCondition() const { return autoModeStopCondition; } // driver-defined
+
+  // Capability flags. Default false; each driver overrides to true for the
+  // fields it actually parses. Consumers should check these before trusting
+  // the corresponding getter.
+  virtual bool hasFlowRate() const { return false; }
+  virtual bool hasBatteryLevel() const { return false; }
+  virtual bool hasScaleTimer() const { return false; }
+  virtual bool hasWeightUnit() const { return false; }
+  virtual bool hasAutoModeStopCondition() const { return false; }
 
   void setWeightUpdatedCallback(void (*callback)(float), bool onlyChanges = false);
   void setLogCallback(LogCallback logCallback) { this->logCallback = logCallback; }
@@ -52,6 +81,16 @@ protected:
   NimBLERemoteService* clientGetService(const NimBLEUUID uuid);
 
   void setWeight(float newWeight);
+
+  // Setters for optional fields. Drivers that parse these call from their
+  // notification handler. Stored centrally so consumers can read via the
+  // public getters without caring which driver the scale is.
+  void setFlowRate(float newFlow) { flowRate = newFlow; }
+  void setBatteryLevel(uint8_t pct) { batteryLevel = pct; }
+  void setScaleTimerMs(uint32_t t) { scaleTimerMs = t; }
+  void setWeightUnit(ScaleWeightUnit u) { weightUnit = u; }
+  void setAutoModeStopCondition(uint8_t c) { autoModeStopCondition = c; }
+
   void log(std::string msgFormat, ...);
   std::string byteArrayToHexString(const uint8_t* byteArray, size_t length);
 
@@ -59,6 +98,11 @@ private:
   using WeightCallback = void (*)(float);
 
   float weight = 0.f;
+  float flowRate = 0.0f;
+  uint8_t batteryLevel = REMOTE_SCALES_BATTERY_UNKNOWN;
+  uint32_t scaleTimerMs = 0;
+  ScaleWeightUnit weightUnit = ScaleWeightUnit::UNKNOWN;
+  uint8_t autoModeStopCondition = 0;
 
   NimBLEClient* client = nullptr;
   DiscoveredDevice device;
--- a/lib/esp-arduino-ble-scales/src/remote_scales.h
+++ b/lib/esp-arduino-ble-scales/src/remote_scales.h
@@ -56,6 +56,7 @@ public:
   virtual bool hasScaleTimer() const { return false; }
   virtual bool hasWeightUnit() const { return false; }
   virtual bool hasAutoModeStopCondition() const { return false; }
+  virtual bool hasTimerControl() const { return false; }
 
   void setWeightUpdatedCallback(void (*callback)(float), bool onlyChanges = false);
   void setLogCallback(LogCallback logCallback) { this->logCallback = logCallback; }
@@ -69,6 +70,10 @@ public:
   virtual bool connect() = 0;
   virtual void disconnect() = 0;
   virtual void update() = 0;
+
+  virtual void startTimer() {}
+  virtual void stopTimer() {}
+  virtual void resetTimer() {}
 
   virtual ~RemoteScales() { clientCleanup(); }
 protected:
--- a/lib/esp-arduino-ble-scales/src/remote_scales.h
+++ b/lib/esp-arduino-ble-scales/src/remote_scales.h
@@ -75,7 +75,15 @@ public:
   virtual void stopTimer()  { /* default no-op; drivers with hasTimerControl()==true override this */ }
   virtual void resetTimer() { /* default no-op; drivers with hasTimerControl()==true override this */ }
 
-  virtual ~RemoteScales() noexcept { clientCleanup(); }
+  virtual ~RemoteScales() noexcept {
+    try {
+      clientCleanup();
+    } catch (...) {
+      // Swallow: destructors must not propagate exceptions (noexcept guarantee).
+      // clientCleanup calls into NimBLE stack which in practice doesn't throw,
+      // but being explicit satisfies the type system and SonarCloud S1048.
+    }
+  }
 protected:
   RemoteScales(const DiscoveredDevice& device);
   const DiscoveredDevice& getDevice() const { return device; }
--- a/lib/esp-arduino-ble-scales/src/remote_scales.h
+++ b/lib/esp-arduino-ble-scales/src/remote_scales.h
@@ -1,0 +1,112 @@
+#pragma once
+#include <NimBLEDevice.h>
+#include <Arduino.h>
+#include <vector>
+#include <memory>
+#include <lru_cache.h>
+
+
+class DiscoveredDevice {
+public:
+  DiscoveredDevice(NimBLEAdvertisedDevice* device) :
+  name(device->getName()), address(device->getAddress()), manufacturerData(device->getManufacturerData()), rssi(device->getRSSI()) {}
+  const std::string& getName() const { return name; }
+  const NimBLEAddress& getAddress() const { return address; }
+  const std::string& getManufacturerData() const { return manufacturerData; }
+  const int getRSSI() const { return rssi; }
+private:
+  std::string name;
+  NimBLEAddress address;
+  std::string manufacturerData;
+  int rssi;
+};
+
+class RemoteScales {
+
+public:
+  using LogCallback = void (*)(std::string);
+
+  float getWeight() const { return weight; }
+
+  void setWeightUpdatedCallback(void (*callback)(float), bool onlyChanges = false);
+  void setLogCallback(LogCallback logCallback) { this->logCallback = logCallback; }
+
+  std::string getDeviceName() const { return device.getName(); }
+  std::string getDeviceAddress() const { return device.getAddress().toString(); }
+  int getRSSI() const { return client != nullptr ? client->getRssi() : 0; }
+
+  virtual bool tare() = 0;
+  virtual bool isConnected() = 0;
+  virtual bool connect() = 0;
+  virtual void disconnect() = 0;
+  virtual void update() = 0;
+
+  ~RemoteScales() { clientCleanup(); }
+protected:
+  RemoteScales(const DiscoveredDevice& device);
+  const DiscoveredDevice& getDevice() const { return device; }
+
+  bool clientConnect();
+  void clientCleanup();
+  bool clientIsConnected();
+  NimBLERemoteService* clientGetService(const NimBLEUUID uuid);
+
+  void setWeight(float newWeight);
+  void log(std::string msgFormat, ...);
+  std::string byteArrayToHexString(const uint8_t* byteArray, size_t length);
+
+private:
+  using WeightCallback = void (*)(float);
+
+  float weight = 0.f;
+
+  NimBLEClient* client = nullptr;
+  DiscoveredDevice device;
+  LogCallback logCallback = nullptr;
+  WeightCallback weightCallback = nullptr;
+  bool weightCallbackOnlyChanges = false;
+};
+
+// ---------------------------------------------------------------------------------------
+// ---------------------------   RemoteScalesScanner    -----------------------------------
+// ---------------------------------------------------------------------------------------
+class RemoteScalesScanner : public NimBLEAdvertisedDeviceCallbacks {
+private:
+  bool isRunning = false;
+  LRUCache alreadySeenAddresses = LRUCache(100);
+  std::vector<DiscoveredDevice> discoveredScales;
+  void cleanupDiscoveredScales();
+  void onResult(NimBLEAdvertisedDevice* advertisedDevice) override;
+
+public:
+  std::vector<DiscoveredDevice> getDiscoveredScales() { return discoveredScales; }
+
+  void initializeAsyncScan();
+  void stopAsyncScan();
+  void restartAsyncScan();
+  bool isScanRunning() const;
+};
+
+// ---------------------------------------------------------------------------------------
+// ---------------------------   RemoteScalesFactory    ----------------------------------
+// ---------------------------------------------------------------------------------------
+
+// This is a singleton class that is used to create RemoteScales objects from BLEAdvertisedDevice objects.
+class RemoteScalesFactory {
+public:
+  std::unique_ptr<RemoteScales> create(DiscoveredDevice device);
+
+  static RemoteScalesFactory* getInstance() {
+    if (instance == nullptr) {
+      instance = new RemoteScalesFactory();
+    }
+    return instance;
+  }
+
+  RemoteScalesFactory(RemoteScalesFactory& other) = delete;
+  void operator=(const RemoteScalesFactory&) = delete;
+
+private:
+  static RemoteScalesFactory* instance;
+  RemoteScalesFactory() {}  // Private constructor to enforce singleton
+};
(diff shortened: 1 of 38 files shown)


lib/esp-arduino-ble-scales/src/scales/bookoo.cpp
```diff
@@ -0,0 +1,321 @@
...
+bool BookooScales::decodeAndHandleNotification() {
+  // Minimum message length check (20 bytes based on protocol definition)
+  if (dataBuffer.size() < RECEIVE_PROTOCOL_LENGTH) {
+    return false;
+  }
+
+  BookooMessageType messageType = static_cast<BookooMessageType>(dataBuffer[1]);
+  uint8_t productNumber = dataBuffer[0];
+
+  size_t messageLength = dataBuffer.size();
```
salekseev: bug: `messageLength = dataBuffer.size()` eats whole buffer as one frame. Two queued 20-byte notifies → 40-byte "msg" → checksum fails → both dropped. Fix: `messageLength = RECEIVE_PROTOCOL_LENGTH`, slice one frame per call.

lib/esp-arduino-ble-scales/src/scales/bookoo.cpp
```diff
@@ -0,0 +1,321 @@
...
+    // Battery percentage (byte 13).
+    RemoteScales::setBatteryLevel(dataBuffer[13]);
+
+    // Auto-mode stop condition (byte 18) -- only meaningful on Ultra scales
+    // where hasAutoModeStopCondition() returns true. We still store it so an
+    // Ultra-aware subclass (or a future firmware-side model check) can read it.
+    RemoteScales::setAutoModeStopCondition(dataBuffer[18]);
+  }
+  else if (productNumber == 0x03 && messageType == BookooMessageType::SYSTEM) {
+    BookooScales::tare();
```
salekseev: bug: SYSTEM inbound → calls `tare()` → writes 0x0A back → scale re-acks → loop / spurious mid-shot tares. Fix: log and ignore, or parse sub-opcodes. Tare belongs on user action only.

lib/esp-arduino-ble-scales/src/remote_scales.h
```diff
@@ -0,0 +1,156 @@
...
+  std::string getDeviceAddress() const { return device.getAddress().toString(); }
+  int getRSSI() const { return client != nullptr ? client->getRssi() : 0; }
+
+  virtual bool tare() = 0;
+  virtual bool isConnected() = 0;
+  virtual bool connect() = 0;
+  virtual void disconnect() = 0;
+  virtual void update() = 0;
+
+  ~RemoteScales() { clientCleanup(); }
```
salekseev: L73: 🔴 bug: non-virtual dtor, deleted via `unique_ptr<RemoteScales>`. UB; `BookooScales::dataBuffer` and cached chars leak. Fix: `virtual ~RemoteScales()`. Worth upstreaming.

lib/esp-arduino-ble-scales/src/scales/bookoo.cpp
```diff
@@ -0,0 +1,321 @@
...
+  }
+  else {
+    RemoteScales::log("Unknown message type %02X: %s\n", messageType, RemoteScales::byteArrayToHexString(dataBuffer.data(), messageLength).c_str());
+  }
+
+  // Remove processed message from the buffer
+  dataBuffer.erase(dataBuffer.begin(), dataBuffer.end());
+
+  // Return whether there's more data to process
+  return dataBuffer.size() < RECEIVE_PROTOCOL_LENGTH;
```
salekseev: risk: return inverted. Caller loops `while (result)`; this returns true when buffer is **short**. Harmless today (early-exit bails next iter), real bug once L125 slicing fixed — multi-frame batches stop after frame 1. Fix: `return dataBuffer.size() >= RECEIVE_PROTOCOL_LENGTH;`.

src/display/core/Controller.cpp
```diff
@@ -364,6 +364,16 @@ bool Controller::isAutotuning() const { return autotuning; }
...
 
 bool Controller::isVolumetricAvailable() const {
+    // Safety gate: if the BLE scale is physically toggled to ounces, its weight
+    // samples are ~28x off from grams -- volumetric stops would either fire
+    // immediately (target 36g reached at v=1.3 "oz" which is actually 36.8g) or
+    // never fire (target 36g but scale values never exceed a few units). Either
+    // way the shot would be silently wrong. Refuse volumetric routing when the
+    // unit is explicitly ounces. UNKNOWN (driver doesn't parse unit) is treated
+    // as GRAM for backwards compatibility with scales that don't report it.
+    if (BLEScales.hasWeightUnit() && BLEScales.getWeightUnit() == ScaleWeightUnit::OUNCE) {
```
salekseev, line 374: risk: gate covers shot-start only. User flips scale to oz mid-brew → weights drop ~28× → volumetric target never hit → shot runs until time safety. Fix: also check unit in per-sample path (`onVolumetricMeasurement` or `BLEScalePlugin::onMeasurement`), abort volumetric → time fallback + `ESP_LOGW`.

salekseev (commented):
3 bugs, 3 risks, 3 nits. I recommend to fix critical ones before merge / upstream. Merging might be a legal issue with this library being MIT licensed and this project being under CC BY-NC-SA 4.0.

lib/esp-arduino-ble-scales/src/scales/bookoo.cpp
```diff
@@ -0,0 +1,321 @@
...
+    weightCharacteristic->subscribe(true, callback);
+  }
+
+  if (commandCharacteristic->canNotify()) {
+    RemoteScales::log("Registering callback for command characteristic\n");
+    commandCharacteristic->subscribe(true, callback);
+  }
+}
+
+void BookooScales::sendMessage(BookooMessageType msgType, const uint8_t* payload, size_t length, bool waitResponse) {
```
salekseev: risk: `msgType` param unread. Callers pack type byte into `payload` manually; checksum overwrites caller's last byte silently. Fix: either frame inside (`{0x03, msgType, len, ...payload, chk}`) and have callers pass data only, or drop the param.

src/display/plugins/BLEScalePlugin.cpp
```diff
@@ -319,6 +328,44 @@ void BLEScalePlugin::onMeasurement(float value) const {
...
     // Safe to call controller method
     controller->onVolumetricMeasurement(value, VolumetricMeasurementSource::BLUETOOTH);
+
+    // If the scale driver also provides native flow rate (e.g. Bookoo), emit
+    // it on the same tick so consumers get it at the scale's native cadence
+    // (~10 Hz) without having to poll. Controller.onVolumetricMeasurement
+    // updates lastBluetoothMeasurement timestamps as a side effect; we reuse
+    // a lighter path here since flow is not gating shot state.
+    if (scale != nullptr && scale->hasFlowRate() && pluginManager != nullptr) {
+        pluginManager->trigger(
```
salekseev, line 369: nit: flow event fires every 10 Hz tick incl. unchanged values; battery/unit already use emit-on-change. Mirror it: cache `lastFlowRate`, compare with epsilon ~0.01, reset in `disconnect()` next to other caches.

src/display/plugins/BLEScalePlugin.cpp
```diff
@@ -319,6 +328,44 @@ void BLEScalePlugin::onMeasurement(float value) const {
...
+            "controller:volumetric-measurement:scale-flow:change",
+            "value", scale->getFlowRate());
+    }
+}
+
+void BLEScalePlugin::pollScaleMetadata() {
```
salekseev, line 375: nit: called every `update()` tick (~100 ms). Early-exit keeps it cheap but it hits 4 virtual getters 10×/s for data that changes at 1 Hz at most. Gate on `lastMetadataPollMs` at 1 Hz.

lib/esp-arduino-ble-scales/src/scales/bookoo.cpp
```diff
@@ -0,0 +1,321 @@
...
+    return false;
+  }
+  subscribeToNotifications();
+  RemoteScales::setWeight(0.f);
+
+  // Disable the scale-side flow-smoothing EMA so consumers see raw per-sample
+  // flow in getFlowRate(). Firmware-side code (ShotHistoryPlugin,
+  // VolumetricRateCalculator) is free to filter if needed; running both EMAs
+  // compounds lag without adding accuracy.
+  disableScaleSmoothing();
```
salekseev, line 58: nit: one-shot, no verify. Dropped BLE write or reconnect via `markedForReconnection` path → scale-side EMA silently re-enabled → double-filter lag returns. Fix: re-issue every Nth `sendHeartbeat`, or parse byte 17 (flow-smoothing switch) and resend on `1`.

thetimetraveler:
Thanks for the thorough review. Addressed all 6 real bugs across two PRs; the three nits are deferred.

**Driver fixes — now upstream** as [gaggimate/esp-arduino-ble-scales#29](https://github.com/gaggimate/esp-arduino-ble-scales/pull/29). Each commit there addresses one of your bullets:

- **Bug 1** (messageLength framing) + **Risk 4** (inverted return) — collapsed into [one commit](https://github.com/jniebuhr/gaggimate/pull/671/commits/56181186): messageLength = RECEIVE_PROTOCOL_LENGTH, success-path erase uses `begin() + messageLength` to preserve queued frames, and the decode-loop sentinel flipped to `>=`. You were right that the two are tightly coupled — a short-buffer `true` with a fixed slicer is exactly the regression.
- **Bug 2** (SYSTEM → tare feedback loop): [`903eb31a`](https://github.com/jniebuhr/gaggimate/pull/671/commits/903eb31a). Replaced the `tare()` call with the log-and-ignore pattern used for unknown message types.
- **Bug 3** (non-virtual dtor): [`1d1413cb`](https://github.com/jniebuhr/gaggimate/pull/671/commits/1d1413cb). `virtual` added. This one is base-class correctness independent of the Bookoo driver — glad you caught it.
- **Risk 6** (dead `msgType` param): [`521f3d57`](https://github.com/jniebuhr/gaggimate/pull/671/commits/521f3d57). Dropped the parameter entirely and documented the checksum-overwrite behaviour so future callers know to reserve `payload[length-1]`. `BookooMessageType` enum stays since it's still used for decoding inbound messages.

**Firmware-only fix — this PR**:

- **Risk 5** (oz guard shot-start only): [`7361a5c6`](https://github.com/jniebuhr/gaggimate/pull/671/commits/7361a5c6). Per-sample guard in `BLEScalePlugin::onMeasurement` clears `volumetricOverride` when the scale reports OUNCE mid-brew, with a latched `ESP_LOGW` to avoid 10 Hz log spam. `Controller::isVolumetricAvailable()` was already the gate consumers check, so the remainder of the shot falls through to time-based stop without a separate path.

**Nits deferred**: the flow-emit-on-change / 1 Hz metadata poll / smoothing resend-on-reconnect ones are all real and I agree with them, but they're event-cadence / plugin-polling concerns that are cleaner as a focused follow-up than bundled here.

**On the license concern**: fair flag. Upstream PR #29 moves the driver changes into the MIT repo where they belong; the `lib/esp-arduino-ble-scales/` tree here is vendored transitionally so this firmware PR stays testable while upstream is in review. Once #29 merges I'll delete the vendored copy and point `platformio.ini` back at the library. Updated `VENDORING.md` ([`41b99e84`](https://github.com/jniebuhr/gaggimate/pull/671/commits/41b99e84)) explains the transition.

Thanks again — really useful review.

thetimetraveler:
Pushed **bce7c0fa** to address a regression introduced by 9e1ec2a5 (the earlier drop of the manual CCCD write in favor of NimBLE's `subscribe(true, ...)`).

**Symptom**: after that commit, the Bookoo's GATT link comes up (`scale->isConnected()` returns true, `/api/scales/info` shows `connected: true`), but no weight notifications ever arrive. The display's brew screen shows no live weight, `isBluetoothScaleHealthy()` falls through the grace period, `isVolumetricAvailable()` returns false, and volumetric targets silently fall back to time-based stops.

**Cause**: the order of operations in `BookooScales::connect()` was:

1. `performConnectionHandshake()` → `sendNotificationRequest()` writes the start-notifications command to the command characteristic.
2. `subscribeToNotifications()` → NimBLE writes CCCD (0x0001) to the weight characteristic.

Bookoo scales require CCCD to be enabled before the start command is processed; receiving the command with CCCD still off causes the scale to silently never begin emitting weight notifications. The previous manual CCCD write in `performConnectionHandshake()` — despite its buggy big-endian bytes — happened before `sendNotificationRequest()` and was enough to put the scale in the right state. Removing it left Bookoo specifically affected because no other driver here sends a pre-subscribe start command.

**Fix**: move `sendNotificationRequest()` out of `performConnectionHandshake()` and into `connect()` after `subscribeToNotifications()` runs. Order becomes:

1. Handshake (characteristics only, no commands)
2. Subscribe (CCCD enabled via NimBLE)
3. Start-notifications command (scale now knows client wants notifications)

Verified on a Bookoo Themis Ultra — live weight resumes showing up on the brew screen, volumetric targets engage.

thetimetraveler:
Added **ff14bd65** as belt-and-suspenders on top of bce7c0fa: an explicit manual CCCD write (LE `{0x01, 0x00}` = 0x0001) to both the weight and command characteristics before `subscribe(true, ...)`. 

Motivation: [NimBLE-Arduino #340](https://github.com/h2zero/NimBLE-Arduino/issues/340) — `subscribe()` can return true without actually writing CCCD on some ESP32 builds, which would leave the scale silent even with the correct ordering. Writing 0x2902 ourselves is idempotent (NimBLE may also write the same value; peripheral sees a dedup) and guarantees the peripheral is in the notifying state regardless of NimBLE's internal path. `subscribe()` is still called to wire the callback on our side.

Verified build passes on the fork. Will report empirical result from hardware after OTA.

thetimetraveler:
Pushed 4 more commits on top of ff14bd65. Summary (top to bottom):

**24309bd7 — `debug(scales): surface scale-side flow-smoothing state`**
Parses byte 17 of the weight notification (flow-smoothing switch) and exposes it via a new virtual `RemoteScales::isFlowSmoothingOn()` overridden in `BookooScales`. Lets consumers verify the `0x08 0x00` command issued in `connect()` was actually honored by the scale. Default impl on the base class returns `false` so non-Bookoo drivers keep current behavior.

**0165f09c — `fix(remote_scales): actually add isFlowSmoothingOn virtual`**
Adds the `virtual bool isFlowSmoothingOn() const { return false; }` declaration to `remote_scales.h` that 24309bd7 expected. (Was missed in a prior edit; separate commit to keep the diff readable.)

**57075f40 — `feat(bookoo): discriminate Ultra vs Mini by advertising name`**
Stores an `isUltra_` flag at construction based on the `BOOKOO_SC_U…` prefix in the advertising name. Gates `hasAutoModeStopCondition()` on it so Minis don't claim byte-18 data they never emit. Adds two Ultra-only commands that no-op on Mini: `setAutoModeStopConditionOnScale(bool onContainerRemoval)` (cmd `0x0B`) and `calibrate()` (cmd `0x09`). Unknown advertising names default to Mini-compatible to avoid sending Ultra-only commands to an unknown peer.

**f1a01644 — `fix(bookoo): always report GRAM unit — value is always grams per spec`** ⭐️
Core correctness fix, verified empirically against a BOOKOO_SC_U. Per the [Ultra protocol spec](https://github.com/BooKooCode/OpenSource/blob/main/bookoo_ultra_scale/protocols.md): *\"The weight value returned in the data packet is always in grams\"*. Byte 5 is the LCD display-unit indicator, not the unit of the transmitted value. The prior code fed byte 5 into `setWeightUnit()`, which `Controller::isVolumetricAvailable()` then used as a guard (\"refuse volumetric if scale is in oz\"). With the LCD in grams, byte 5 is `0x01` (\"displays oz\" semantically but in the LCD/control sense) → driver marked unit as OUNCE → volumetric silently refused → brew-by-weight never engaged on a Bookoo Ultra with the LCD in grams. Switching the LCD to oz inverted byte 5 and accidentally made it work. Right fix: `BookooScales` always reports `GRAM` from `getWeightUnit()` since the value is always grams regardless of display. Tested on hardware; live weight flows and volumetric target fires correctly independent of the LCD state.

jniebuhr:
Integrated manually

PR #676, opened 2026-04-19
feat: Replace HomeKit with Matter (esp_matter 1.4.0, dual-framework)

salekseev:
## Summary

- **Drop HomeSpan (HomeKit-only)** — removes `HomekitPlugin.{h,cpp}` and `homespan/HomeSpan@1.9.1`
- **Add Matter over WiFi** via `esp_matter@1.4.0` (pulled from the IDF Component Manager registry) as a managed component, running under pioarduino 55.x dual-framework (`framework = arduino, espidf`)
- **BLECoordinator** singleton serialises NimBLE ownership between Matter commissioning BLE and the BLE scale central
- **MatterPlugin** bridges GaggiMate events ↔ Matter clusters: `OnOff` (brew/standby), `TemperatureControl` (boiler target), `TemperatureMeasurement` (current temp), gated behind `-DGAGGIMATE_MATTER`
- **Platform upgrade** to pioarduino 55.03.38-1 (Arduino core 3.3.8 / IDF 5.5.4) across all envs; NimBLE-Arduino bumped to 2.x; `gaggimate-ble-scales` vendored with NimBLE 2.x patches
- **Partition tables**: `partitions/matter_16mb.csv` (6 MB dual-OTA + `fctry` + `matter_kvs` + 3.5 MB SPIFFS), `partitions/matter_8mb.csv` (6 MB single-slot no-OTA — 5.5 MB firmware × 2 OTA won't fit 8 MB)

## Status

**Mostly functional.** Hardware-verified on LilyGo T-RGB (16 MB / 8 MB OPI PSRAM):

- ✅ Matter commissioning from **Apple Home** via BLE → fabric join over WiFi → mDNS operational
- ✅ `OnOff` cluster: Apple toggle round-trips to `Controller::activateStandby` / `deactivateStandby`
- ✅ `TemperatureControl::TemperatureSetpoint`: Apple setpoint writes land in `Controller::setTargetTemp`
- ✅ `TemperatureMeasurement::MeasuredValue`: current boiler temp streams to Apple every ~250 ms
- ✅ Existing features survive pioarduino upgrade: LVGL UI, WiFi AP + STA, BLE scale central (Bookoo tested), SPIFFS web UI, OTA (legacy single-slot), WebSocket, settings NVS, MQTT

**Not yet tested:** Google Home / Alexa / SmartThings / HA Matter controllers; 24 h soak; OTA with Matter state in `matter_kvs`.

## ⚠️ Temporary: `esp-arduino-ble-scales` pinned to open upstream PR

`platformio.ini` pulls the scales lib from [gaggimate/esp-arduino-ble-scales#28](https://github.com/gaggimate/esp-arduino-ble-scales/pull/28) (branch `salekseev:feat/pioarduino-support`) rather than upstream `main`. PR 28 adds pioarduino / Arduino ESP32 3.x build support + NimBLE 2.x port, which this tree needs for BLE coexistence with the esp-matter commissioning stack.

**Before this PR merges**, flip `platformio.ini` back to the canonical upstream URL once PR 28 lands upstream:

```diff
-    https://github.com/salekseev/esp-arduino-ble-scales.git#feat/pioarduino-support
+    https://github.com/gaggimate/esp-arduino-ble-scales
```

Earlier revisions of this PR included a vendored copy under `lib/gaggimate-ble-scales/`; that was rebased out in favour of the upstream-PR pin.

## The dep-graph detective work (why `esp_matter@1.4.0` and no other version)

arduino-esp32 3.3.x pins `esp_insights=1.2.2` exact. `esp_matter@1.4.2+` and `release/v1.5` bump that to `^1.2.4`, breaking the IDF Component Manager resolver. **Only `esp_matter@1.4.0` from the registry intersects** — it's the last release whose transitive pins match arduino-esp32 3.3.x.

Phase 5a (PlatformIO dual-framework with esp-matter git submodule) and Phase 5b (source-built via Component Manager override) both hit the same graph and stalled. Phase 5c was an idf.py spike that reconfirmed both build systems use the same resolver. Phase 5d discovered the 1.4.0 pin works for both. Full recipe in `sdkconfig.matter.defaults` header.

## Critical build-system bits

- `src/idf_component.yml` pins `espressif/esp_matter: "1.4.0"`.
- `CMakeLists.txt` at project root sets CHIP CXX flags (`-std=gnu++17;-DCHIP_HAVE_CONFIG_H;-Wno-overloaded-virtual`).
- `src/CMakeLists.txt` globs `src/display/**`; drivers/ui excluded when `-DGAGGIMATE_HEADLESS_BUILD=1` (headless envs that don't pull TFT_eSPI / GFX Library / LVGL).
- `scripts/patch_arduino_libs.py` strips `ESP_SR / Zigbee / BluetoothSerial / Matter / Insights / RainMaker / OpenThread / SimpleBLE / ESP_HostedOTA / HTTPClient / HTTPUpdate / ArduinoOTA / WebServer` from `framework-arduinoespressif32/CMakeLists.txt` via `list(REMOVE_ITEM ARDUINO_ALL_LIBRARIES ...)`. Kconfig `ARDUINO_SELECTIVE_*` disables are unreliable under IDF 5.5. Also adds the GCC 14 `HTTPClient::RequestArgument(String, String)` ctor workaround.
- `board_build.cmake_extra_args = -DEXECUTABLE_COMPONENT_NAME=src ...` — esp_matter's CMakeLists looks for a component named `main` by default; PlatformIO's app component is named after `PROJECT_SRC_DIR` basename (= `src`).

## Build matrix

| env | flash | framework | Matter | partitions | notes |
|---|---|---|---|---|---|
| `display` | 16 MB | arduino+espidf | yes | `partitions/matter_16mb.csv` | dual-OTA |
| `display-headless` | 16 MB | arduino+espidf | yes | `partitions/matter_16mb.csv` | dual-OTA, no UI |
| `display-headless-8m` | 8 MB | arduino+espidf | yes | `partitions/matter_8mb.csv` | single-slot no-OTA |
| `display-headless-4m` | 4 MB | arduino | no | `max_app_4MB.csv` | Matter won't fit |
| `controller` | 8 MB | arduino | n/a | board default | consistency |

## Runtime fixes bundled

- **MatterPlugin reentrancy guard** (`applyingFromMatter` flag around `OnOff` + `TemperatureSetpoint` writes): breaks Matter-write → Controller event → `onXChange` → Matter update → pre-attribute-change callback → Matter write recursion. Both paths guarded.
- **BLEScalePlugin 2 s retry gate**: prevents tight-loop flood when target MAC isn't in the scan cache.
- **DefaultUI `lv_obj_is_valid` guard**: protects against LVGL anim paths that free the old screen asynchronously before `lv_obj_del`.
- **`CONFIG_BT_NIMBLE_MAX_CONNECTIONS=4`**: enough slots for CHIPoBLE advertiser + inbound commissioner + scale central + headroom.
- **`CONFIG_ASYNC_TCP_STACK_SIZE=16384`** (was 4096): IDF 5.5 frames overflowed 4 KB on profiles JSON parse + SPIFFS read.

## pioarduino toolchain compat (not Matter-specific)

- `LOG_TAG`: `const String` → `const char *` — IDF 5.5 `esp_log` drops the Arduino `String` overload.
- `xTaskHandle` → `TaskHandle_t` across `lib/**` and `src/display/**` — legacy alias removed in newer FreeRTOS.
- `Driver.h` virtuals inline — linker missed the vtable without a `.cpp`.
- `utils.h` explicit `<cstdio>/<stdexcept>/<string>/<vector>/<functional>` — Arduino.h transitive include reduction under Core 3.3.x.
- `[[noreturn]]` on `Max31855Thermocouple::monitorTask` + `PressureSensor::loopTask` declarations — GCC 14 rejects impl/decl mismatch.
- `GitHubOTA.cpp` log param names aligned with ctor signature (param-name check now error).

## Test plan

- [x] Commission from **Apple Home** via QR / pairing code
- [x] OnOff round-trip (Apple → brew/standby)
- [x] TemperatureSetpoint round-trip (Apple → boiler target)
- [x] TemperatureMeasurement stream (boiler current → Apple)
- [x] BLE scale reconnect after Matter commissioning
- [x] Web UI loads from SPIFFS (profiles, settings)
- [ ] Commission from **Google Home** / **Alexa** / **SmartThings** / **Home Assistant Matter**
- [ ] BLE coexistence soak: commission / decommission / re-commission while scale streams
- [ ] OTA update with `matter_kvs` fabric preserved
- [ ] 24 h soak: Matter + scale + UI + MQTT concurrent
- [ ] Regression: LVGL UI stability, WiFi reprovisioning, WebSocket

Co-Authored-By: Claude Code](https://claude.com/claude-code)


platformio.ini
```diff
@@ -88,8 +90,12 @@ board = seeed_xiao_esp32s3
 [env:display-headless-4m]
 extends = env:display-headless
 board = esp32-s3-supermini
+build_flags =
+    ${display_common.build_flags}
+    -DGAGGIMATE_HEADLESS
 
 [env:controller]
+platform = espressif32@6.12.0
```
jniebuhr: Controller should be running on the same ESP-IDF version

platformio.ini
```diff
@@ -99,7 +105,7 @@ lib_deps =
 	FS
 	SPIFFS
 	ble_ota_dfu
-	h2zero/NimBLE-Arduino@^1.4.0
+	h2zero/NimBLE-Arduino@^2.3.0
```
jniebuhr, line 139: please test BLE OTA features between the two nimble versions, this caused problems in the past.

old -> new, new -> old, new -> new

src/display/drivers/AmoledDisplayDriver.cpp
```diff
@@ -60,9 +60,23 @@ bool AmoledDisplayDriver::supportsSDCard() { return true; }
 bool AmoledDisplayDriver::installSDCard() { return panel->installSD(); }
 
 bool AmoledDisplayDriver::testHw(AmoledHwConfig hwConfig) {
-    // No Wire on these pins, definitely wrong board
-    if (!Wire.begin(hwConfig.i2c_sda, hwConfig.i2c_scl))
+    // Cold-boot peripherals need a moment; also recover from any prior
```
jniebuhr, line 63: this feels unrelated to the PR

src/display/drivers/LilyGo-T-RGB/LilyGo_RGBPanel.cpp
```diff
@@ -10,7 +10,6 @@
 #include "LilyGo_RGBPanel.h"
 #include "utilities.h"
 #include <display/drivers/common/RGBPanelInit.h>
-#include <esp_adc_cal.h>
```
jniebuhr, line 13: feels unrelated to this PR

src/display/core/Controller.cpp
```diff
@@ -107,8 +124,20 @@ void Controller::connect() {
     connectStartTime = millis();
     pluginManager->trigger("controller:startup");
 
-    setupWifi();
+    ESP_LOGI("HEAP", "pre-BT   internal=%u largest=%u total=%u",
```
jniebuhr: I recently added a helper method for this

lib/ble_ota_dfu/src/ble_ota_dfu.cpp
```diff
@@ -150,29 +150,7 @@ uint16_t BLEOverTheAirDeviceFirmwareUpdate::write_binary(fs::FS *file_system, co
     }
 }
 
-void BLEOverTheAirDeviceFirmwareUpdate::onNotify(BLECharacteristic *pCharacteristic) {
```
jniebuhr, line 153: why was this removed?

jniebuhr (changes requested):
In my research it seemed Matter wouldn't work without certification, did you try it out?Also I'd prefer not to vendor the scales lib just now, we already had a PR for that to upgrade it to NimBLE 2.x, the lib is under our control

platformio.ini
salekseev: Done in commit 989eeed7 — controller env now inherits the top-level `[env]` platform (pioarduino 55.03.38-1), same Arduino Core 3.3.8 + IDF 5.5.4 as display. Required deleting the stale `~/.platformio/packages/framework-arduinoespressif32` (Core 2.0.17 stub) to force a redownload of the 3.3.8 core, plus a couple of source fixes for GCC-14 strictness (`[[noreturn]]` header/impl match on `Max31855Thermocouple::monitorTask` + `PressureSensor::loopTask`). Controller builds clean on the same stack now.

platformio.ini
salekseev: Test matrix acknowledged. Currently positioned for **new → old** (display on this branch = NimBLE 2.5.0, controller still on v1.8.0-11-gf7c186b2 = master = NimBLE 1.x). Normal BLE comm already verified working across the version boundary (scale central, temp streaming, mode/profile sync), but OTA is a separate GATT code path so it needs dedicated exercise.

Plan:
1. Trigger controller OTA from new display → old controller via web UI, confirm chunk transport + post-reboot version bump + clean re-advertise/reconnect.
2. Flash new controller, repeat OTA flow end-to-end for **new → new**.
3. Checkout master on display env, flash → push OTA to new controller, for **old → new**. (Easiest path since I already have a stock master display build artifact I can revert to.)

Things to watch specifically across the 1.x↔2.x boundary: MTU negotiation (different defaults), CCCD subscribe timing, and the disconnect-on-reboot race where the 2.x side can re-scan faster than the 1.x peer re-advertises. Will paste logs when runs complete.

src/display/drivers/AmoledDisplayDriver.cpp
salekseev: Related — though indirectly, not a Matter-specific change. Under pioarduino core 3.x the new Arduino I2C driver (`hal-i2c-ng`) is stricter about re-init: calling `Wire.begin()` when an earlier attempt left the peripheral in `ESP_ERR_INVALID_STATE` just fails silently and `testHw` returned false during cold-boot probing. Two fixes required:

1. `Wire.end()` + 20 ms before re-begin to clear the stuck state.
2. Assert the `lcd_en` power rail (shared by LCD + touch on LilyGo T-Display-S3 AMOLED) *before* I2C probe — on a cold boot the touch chip doesn't ACK until powered, and without a touch ACK the whole display probe bails.

Without either, auto-detection loop panics with "No compatible display driver found" on first flash. Could split into a separate commit + PR if you want — happy to move it out of this series.

src/display/drivers/LilyGo-T-RGB/LilyGo_RGBPanel.cpp
salekseev: Required by the pioarduino / IDF 5.5 bump. `esp_adc_cal.h` and the `esp_adc_cal_characterize` / `esp_adc_cal_raw_to_voltage` API were deprecated in IDF 4.x and **removed** in IDF 5.x — code referencing them doesn't compile under pioarduino core 3.3.x. The replacement is `analogReadMilliVolts()` (in the Arduino Core `esp32-hal-adc` wrapper), which performs the calibration internally using eFuse-calibrated curves. Net result is functionally identical (averaged millivolt readout, then × 2 for the divider) with fewer moving parts.

Happy to split into a prep commit if you'd prefer it reviewed separately, but it's a hard prereq for the platform bump — the file won't link otherwise.

src/display/core/Controller.cpp
salekseev: Good catch — you mean `measure_heap(label, callback)` in `src/display/core/utils.cpp`? The raw `heap_caps_get_*` calls here were added while triaging a DRAM shortfall during Matter init (PSRAM was eating \~180 KB of static data and we needed before/after at each init step to find the leak). I'll refactor to use your helper in a follow-up commit — something like:

```cpp
measure_heap("setupBluetooth", [this]{ setupBluetooth(); });
measure_heap("setupWifi", [this]{ setupWifi(); });
```

Keeps the observability during Matter bring-up without duplicating the logging pattern. Will push that + the `onNotify` no-op restore below in the same cleanup commit.

lib/ble_ota_dfu/src/ble_ota_dfu.cpp
salekseev: Two factors — happy to restore if you'd rather keep the symmetry with `onWrite`:

1. **Body was already a no-op in every shipped build.** The entire function body was wrapped in `#ifdef DEBUG_BLE_OTA_DFU_TX`, and `DEBUG_BLE_OTA_DFU_TX` is never `#define`d anywhere in the tree (checked with `git grep DEBUG_BLE_OTA_DFU_TX master` — only the `#ifdef` site exists, no matching `#define` / build flag / env config). So the override compiled to empty regardless.

2. **NimBLE 2.x signature changed.** The 2.x `NimBLECharacteristicCallbacks::onNotify` has a different signature — keeping the 1.x signature breaks the `override` specifier. Migrating it meant rewriting an empty function against a new interface, so I deleted instead.

If you want it back for symmetry, I'll restore it as a 2.x-signature no-op with a comment explaining it's reserved for future debug tracing. Just confirm and I'll add it to the cleanup commit alongside the `measure_heap` refactor.

salekseev:
**Note on `WebUIPlugin.cpp` — `setCloseClientOnQueueFull(false)`**

Hit this crash during long soak of the pioarduino / Matter branch:

```
E async_ws: [/ws][N] Too many messages queued: closing connection
Guru Meditation Error: Core 0 panic'ed (LoadProhibited). EXCVADDR 0x14.
PC: std::_Function_handler<void (), AsyncMiddlewareChain::_runChain(...)>::_M_manager
#2 AsyncMiddlewareChain::_runChain at Middleware.cpp:56
```

Root cause is in ESPAsyncWebServer (not in our code): with `closeWhenFull = true`, `AsyncWebSocketClient::_queueMessage` calls `_client->close()` synchronously when the WS queue fills, which runs `_onDisconnect → _handleDisconnect → ~AsyncWebSocketClient` on the same stack. If any HTTP request on the same `AsyncClient` is mid-flight inside the nested `AsyncMiddlewareChain::_runChain` (WebRequest.cpp:851–865), the stack-local `next` `std::function` it captures by reference gets torn down under it — the outer `next()` then reads a zero `_M_manager` at offset `0x14`.

Filed upstream: ESP32Async/ESPAsyncWebServer#433 (with suggested deferred-close fix). PR #414 / #427 / #429 locked the WS side but don't cover this cross-subsystem reentrancy.

Workaround in this PR is one line: `client->setCloseClientOnQueueFull(false)` at `WS_EVT_CONNECT`. Excess frames are silently dropped instead of destroying the client; our `STATUS_PERIOD` rate limit keeps queue growth bounded. Can drop the workaround once upstream lands a deferred-close fix.

lib/ble_ota_dfu/src/ble_ota_dfu.cpp
salekseev: Restored as a `logTxPacket(BLECharacteristic *)` helper in a891f8ef / 3f803b23 — gated on `DEBUG_BLE_OTA_DFU_TX`, matching the original opt-in behaviour. Can't go back to `onNotify` itself since NimBLE-Arduino 2.x removed it from `NimBLECharacteristicCallbacks`, but the helper runs before every `->notify()` site in this file (both `onWrite` command paths and the three `BLE_OTA_DFU::send_OTA_DFU` overloads) and compiles to a no-op without the flag.

Also rolled back the `onWrite` cognitive-complexity refactor I'd added to clear cpp:S3776 — it was pure diff-bloat. Only the one-liner typo fix (`pData[i]` → `pData[index]` inside the `DEBUG_BLE_OTA_DFU_RX` hex-dump loop) stayed.

salekseev:
Real testing has shown Matter commissioning on feature/matter-migration is blocked by a CHIPoBLE bug in esp-matter 1.4.0 (BTP handshake response not reaching phone — 54 ms post-BTP disconnect from both Apple
 Home and HA; CHIP detail logs are compiled out of the prebuilt lib so we can't fully instrument). Matter 1.4.2 introduces Wi-Fi-only commissioning which would sidestep this whole class
 of BLE bugs; esp-matter hasn't caught up.

 Rather than keep blocking the platform work on a stalled-upstream fight, split:

 - New branch feature/pioarduino-refresh off master: platform/dep refresh only — at separate PR https://github.com/jniebuhr/gaggimate/pull/677.
 - feature/matter-migration stays parked until esp-matter ships Matter 1.4.2 + Wi-Fi-only commissioning (or we rewrite BTP/BLE ourselves, not worth it today).

src/display/drivers/AmoledDisplayDriver.cpp
jniebuhr: Ok sorry, didn't realize

PR #680, opened 2026-04-23
fix(#672): replace PID autotuner with sTune + guard zero-gain writes

### LINKED ISSUE #672: [Bug] PID Autotune gives values of 0,0,0,0
### Brief summary of the bug

Autotune process runs without error but provides bogus values

Gaggimate Pro 1.1, Firmware 1.8.0, Rancilio Silvia v1

The default Silvia PID values from the setup instructions cause a fair amount of heating overshoot (6-7 C at times), so I thought I'd try running the Autotune function. After several attempts, the output values are consistently:

-0.000,-0.000,0.000,0.000

While the Autotune function is running, I can see the temperature (and pressure) rise in the web UI. But after running it, the heater doesn't seem to work again until I restart Gaggimate.

The setup seems to work fine otherwise - I'm getting good temperature readings and  good pressure readings, programs seem to run correctly, etc. while operating the machine. 

Happy to provide further info if needed.

Yes

### Steps to reproduce

Run the PID Autotune function on a cold machine - happens every time.

### Expected behavior

It should provide some useful values

No

```shell

```

### Firmware version

1.8.0, both display and controller

Rancilio Silvia v1

### Additional context

_No response_

salekseev:
## Context

Fixes #672. Reporter (`wixoff`) and I both reproduce on Rancilio Silvia v1: running **PID Autotune** from cold prints `-0.000,-0.000,0.000,0.000` in the web UI, the heater stays dead until reboot, and the persisted PID gets nuked to zeros in NVS. Other subsystems (temp read, pressure) otherwise healthy.

Two independent bugs compound into the observed failure:

### Bug A — crash-class: zero-gain payloads written to live PID + NVS
`lib/NayrodPID/src/Autotune/Autotune.h` declares `float Kp, Ki, Kd, Kff;` with no default initializer. On the 20 s hardcoded timeout path (`Autotune.cpp:48-49`), `finished=true` is set without computing gains → garbage bytes emitted. The display callback at `src/display/core/Controller.cpp:167-175` then writes those straight to `settings.setPid(...)` (persistent NVS) and the live SimplePID receives `setTunings(0,0,0)`. Heater integrator clamps to zero output permanently. Only a power cycle recovers it.

### Bug B — algorithm wrong shape for GaggiMate hardware
GaggiMate uses a K-type thermocouple on a **MAX31855** SPI converter (0.25 °C LSB quantization, per docs.gaggimate.eu). The current detector in `Autotune.cpp` uses a 4-sample least-squares slope compared against fixed `epsilon = 0.1 °C/s` at 1 Hz — below the quantization step, fighting noise instead of signal. It then uses *detector-trigger latency* as "dead time L", never estimates τ, and its phase-margin-over-integrator-plus-delay design (mislabeled "Ziegler-Nichols" in comments) is not a recognized industry method.

On a Silvia v1 (1000 W, 0.3 L brass boiler, shell-mounted probe) the thermocouple lag from boiler-water to external-brass-thread is ~15–30 s, which the 20 s timeout can trip before the curve even inflects. Silvia's thermal mass easily clears 0.1 °C/s *at the water*, but not always *at the external probe* within 20 s — producing the failure shape reporter sees.

## Algorithm comparison — why sTune, and why not the alternatives

Three options evaluated before committing to sTune. Summary: NayrodPID (current) is a custom design with hardware-mismatched defaults; sTune is a textbook open-loop step-test identifier with FOPDT extraction; `rancilio-pid/Arduino-PID-Library` is **not an autotuner at all**, it's a runtime controller fork (scope = SimplePID replacement, not this PR).

### Autotuners: NayrodPID (current) vs sTune (proposed)

| Aspect | NayrodPID Autotune | sTune |
|---|---|---|
| Method | Open-loop step, custom slope detector | Open-loop step, inflection-point (tangent method) |
| Slope threshold | Fixed `epsilon = 0.1 °C/s`, 4-sample LS fit, 1 Hz sample | Adaptive: "averaged input moved beyond one resolution value from start" (`pvAvgRes`) |
| MAX31855 fit | **Sub-LSB** (0.1 °C/s < 0.25 °C step per 1 s) — fires on quantization noise | **LSB-aware** — directly tracks 0.25 °C steps |
| Test duration | Hardcoded 20 s timeout | Configurable `testTimeSec`, ends at inflection (~½τ) |
| Model extracted | None (L ≈ trigger latency; τ never estimated) | Full FOPDT: dead time L, time constant τ, process gain K |
| Tuning rules | Custom phase-margin-over-integrator-plus-delay (mislabeled "ZN" in code) | 10 rules × {P, PI, PID}: Ziegler-Nichols, Damped Oscillation, **No Overshoot**, Cohen-Coon, Mixed |
| Timeout behavior | Emits uninitialized zero gains → nukes NVS + live PID | Returns `sample`/`test` enum; no callback until `tunings` state |
| License | Project-local, single maintainer | MIT, widely deployed, ESP32-tested |

### The Silvia evidence points at No Overshoot

User's hand-tuned bench data on this exact machine (`../../jniebuhr/gaggimate/pid-tuning-summary.csv`, 13 shots) settled on:

> **Kp=112.315, Ki=0.658, Kd=400.0** — 1.0 °C steady-state swing, 17 s recovery to 2.7 °C overshoot, picked explicitly for *lowest overshoot + fastest recovery*.

sTune's `NoOvershoot_PID` rule (Åström-Hägglund variant) computes gains from FOPDT `{L, τ, K}` specifically to suppress overshoot — same optimization target as the empirical shot-13 selection. NayrodPID's design has no equivalent — its 35 % Kd reduction heuristic always biases toward *lower* Kd than the bench data says is right.

### Why not `rancilio-pid/Arduino-PID-Library`?

Fork of Brett Beauregard's Arduino PID v1.2.1. **Not an autotuner** — it's a runtime PID controller, same scope as `lib/NayrodPID/src/SimplePID` (which this PR keeps). Value-adds vs upstream Beauregard:

- EWMA low-pass filter on input before D-term (reduces derivative kick from MAX31855 quantization jumps)
- Separate integrator limits via `SetIntegratorLimits()` (looser than output clamp — lets integrator build past saturation horizon on cold start without output windup)
- Debug getters for P/I/D components

Both features are real improvements for this hardware, but:
1. Orthogonal to #672. Swapping controllers requires retuning every shipped profile.
2. SimplePID + user's shot-13 gains already prove steady-state is solvable with current controller — the bug is in **finding** gains, not **running** them.
3. Filed as potential future work; out of scope here.

## Fix

Replace the custom autotuner with **sTune** (David Lloyd, MIT, v2.4.0) — open-loop inflection-point identifier with FOPDT extraction and 10 selectable tuning rules, packaged for Arduino/PlatformIO. Hold SimplePID as the runtime controller (untouched).

### Concrete wins on this stack

1. **Resolution-aware dead-time detection** addresses MAX31855's 0.25 °C LSB directly.
2. **Configurable test length** lets Silvia's 15–30 s shell-probe dead time play out (default `testTimeSec=90`, inflection ~45–60 s).
3. **NoOvershoot_PID default rule** matches user's empirical preference.
4. **Footprint**: ≈ 20 KB flash, ~6 %×samples×4 B RAM (≈ 120 B @ 250 samples).

### Commits

| # | Commit | Purpose |
|---|---|---|
| a | `deps: add dlloydev/sTune@^2.4.0 to controller env` | `platformio.ini` only |
| b | `autotune: replace NayrodPID Autotune with sTune inflection-point tuner` | `Heater.{h,cpp}` — swap autotuner, non-blocking state machine, emergency stop via `tuner->SetEmergencyStop(MAX_AUTOTUNE_TEMP)` |
| c | `autotune: guard against zero/NaN result payloads` | `Controller.cpp` — reject invalid gains before NVS write; fire `controller:autotune:failed` |
| d | `autotune: retune UI knobs for sTune + failed event` | Web UI + `WebUIPlugin` — relabel "Tuning Goal"→"Test Duration" (default 90 s, range 30–300), "Window Size"→"Sample Count" (default 250, range 200–500); surface failed state distinctly |
| e | `autotune: remove NayrodPID Autotune module` | Delete unused `Autotune.{cpp,h}` |

The BLE wire format is unchanged (`req:autotune-start` still carries `time` and `samples` ints) — back-compatible with older displays. Field *semantics* change: `time` now carries sTune's `testTimeSec` (was an unrelated aggressiveness knob).

## Flash size delta

| Env | After | Notes |
|---|---|---|
| controller | 682 169 B / 3 342 336 B (20.4 %) | +~20 KB for sTune — within budget |
| display | 4 887 757 B / 6 553 600 B (74.6 %) | negligible (guard + UI only) |

RAM: controller 10.1 % (33 KB / 320 KB).

## Test plan

- [ ] Cold-start Silvia v1, default knobs (`testTimeSec=90, samples=250, NoOvershoot_PID`). Expect: test ends at inflection ~45–60 s, sensible Kp/Ki/Kd emitted, heater resumes heating, NVS updated, subsequent brew holds setpoint ±1 °C.
- [ ] Failure path: force timeout (`testTimeSec=30` on a slow boiler). Verify UI shows "Autotune Failed", heater resumes with previously-loaded gains, NVS untouched.
- [ ] Overtemp path: `tuner->SetEmergencyStop(125 °C)` trips. Verify heater drops to 0, no zero-gain callback fired, existing gains preserved.
- [ ] Rule comparison: run with `ZN_PID`, `DampedOsc_PID`, `NoOvershoot_PID` on warm Silvia. Compare steady-state swing + recovery against `pid-tuning-summary.csv` shot-13 targets (≈1 °C swing, 17 s recovery). Expect `NoOvershoot_PID` closest.
- [ ] Regression: manually restore shot-13 (`112.315, 0.658, 400, 0`), brew one shot, confirm baseline unchanged.
- [ ] Cross-machine: run on Gaggia Classic with `testTimeSec=60`, verify gains emitted.

## Immediate workaround for affected users (no firmware change)

Enter known-good values manually via **Web UI → Settings → PID**:
- Rancilio Silvia v1: `Kp=112.315, Ki=0.658, Kd=400.0, Kf=0` (from `pid-tuning-summary.csv` shot-13)

---

Draft until bench validation on Silvia completes. Will update the test plan with measured values before marking ready.

Co-Authored-By: [Claude Code](https://claude.com/claude-code)


salekseev:
Superseded by #683. #683 replaces the sTune approach with Skogestad's SIMC-for-integrator+lag rule — on-bench testing showed every FOPDT rule sTune ships structurally caps Kd near 1 on this hardware, while SIMC reproduces the hand-tuned Kd≈400 regime and ships a 0.4 °C steady-state swing (vs sTune's 4.5 °C and hand-tuned's 1.0 °C). All #672 crash-class guards from this PR are preserved in #683. Closing in favor of #683.

PR #684, opened 2026-04-25
feat: Add pump flow calibration to Settings UI

ggazzo:
Adds a "Calibrate" button next to the Pump Flow Coefficients field in
Machine Settings. 
image

Opens a modal that runs the standard 1 bar / 9 bar
measurement profile, parses the resulting .slog, computes new
coefficients (real flow / estimated flow factor multiplied by current
coefficients), and offers to write them back via /api/settings.
image

Behind the scenes, add an auxiliary profile, pulls the shot and analyze the data. The analysis script is exactly the same as suggested in the Discord channel: https://discord.com/channels/951416527721230336/1410049551791951943

image

## Summary by CodeRabbit

* **New Features**
  * Pump Flow Calibration modal added to Settings with guided workflow, start/apply controls, and guarded close behavior.
  * Live, scrolling calibration logs and phase-specific header/footer controls.
  * Idle view with instructions and warnings, plus primary/secondary buttons.
  * Automated analysis showing 1‑bar and 9‑bar results, coefficient comparisons, and ability to save new pump coefficients.
* **Utilities**
  * New calibration analysis and coefficient parsing utilities.


thetimetraveler:
Math checks out — 2-pt fit against linear `pumpModelCoeffs`, and `/api/settings` POST is partial in `WebUIPlugin.cpp:461-477` so writing only the one field is safe. A few things worth tightening:

**Silent failure when no scale is paired.** `analyze()` reads `(last.v - first.v) / time`, and `v` decodes as scale weight (`V: 9, // volumetric weight`, `WEIGHT_SCALE=10` in `parseBinaryShot.js`). With no Bluetooth scale connected, every `v` is 0 → `actualFlow=0` → passes the existing `Number.isFinite(actualFlow)` guard → `factor=0` → "Save to machine" writes `"0.000,0.000"` and the pump model is broken. Probably want both: gate Start on `systemInfo.bluetoothScaleConnected`, and add an `actualFlow > 0` guard alongside the existing `estimatedFlow > 0`.

**No plausibility bound on the factor.** A user who only manages to reach ~5 bar instead of 9 bar (valve too open) still gets a "result" they can apply. A sanity range — say `0.5 < factor < 2.0` — would catch most failure modes that don't surface as a thrown error.

**Calibration profile is left selected after the run.** `req:profiles:save` + `req:profiles:select` make it active; the user's previous profile isn't restored, so they walk away from calibration with the wrong profile selected. Either snapshot+restore the prior selection on close, or avoid persisting the calibration profile at all.

Minor: `SLOG_HEADER_MIN = 128` is the v4 header size; v5+ headers are 512 bytes, so an empty v5 `.slog` would pass the readiness check at 128 and parse as 0 samples. Probably worth gating on the version-appropriate threshold (or just on `samples.length > 0`).

CodeRabbit's open timer-leak finding is also still real — `safetyId` in `waitForShotEnd` isn't cleared by `reset()` or `detachStatusListener()`, so a discard mid-run leaves a stale 5-min reject that fires into a subsequent run's state.

PR #687, opened 2026-05-02
fix: Split Kff out of PID after save round-trip

crwper:
The settings form splits the firmware's `Kp,Ki,Kd,Kff` CSV into two
inputs on initial load (`formData.pid` = first three fields,
`formData.kf` = the fourth) and recombines them on submit. The
post-save response handler was missing the same split, so after
clicking Save the response's full CSV ended up in `formData.pid`
unchanged — the PID input visibly showed four numbers instead of
three, and the next Save would build `${formData.pid},${formData.kf}`
into a five-field string. The split-on-load logic only reads the
fourth field, so Kff still survived round-trips, but the stored CSV
gained a redundant trailing field on every successive save without a
page refresh.

Fix: extract the existing inline split into a small `splitPidString`
helper and apply it in both the initial-load `useEffect` and the
post-save merge in `onSubmit`.

Verified by `npm run lint:check` and `npm run build`. No firmware
changes; the fix is entirely in `web/src/pages/Settings/index.jsx`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed settings corruption where firmware PID fields could grow and combine incorrectly across repeated saves.
  * Ensured feedforward (KF) is populated with a default when missing so PID and KF remain stable and consistent after save/load cycles.


crwper:
Amended to add a JSDoc on splitPidString to satisfy the docstring coverage check.

jniebuhr:
@cla-bot check

PR #690, opened 2026-05-03
Fix: flush pending settings to NVS before restart

crwper:
Save() defers writes by up to 5 seconds via a background task. When the settings form is submitted with the restart flag, ESP.restart() fires immediately after the response is sent, so any setting changed in that POST is lost if the 5-second flush window hasn't elapsed.

This PR forces a synchronous save before restarting so requested changes survive the reboot.

## Summary by CodeRabbit

* **Bug Fixes**
  * Settings are now saved to persistent storage before triggering a restart from the web interface, preventing loss of configuration.

* **Documentation**
  * Clarified web settings endpoint behavior and restart/save semantics in inline documentation.

### DIFF
--- a/src/display/plugins/WebUIPlugin.cpp
+++ b/src/display/plugins/WebUIPlugin.cpp
@@ -458,6 +458,27 @@ void WebUIPlugin::handleProfileRequest(uint32_t clientId, JsonDocument &request)
     ws.text(clientId, buffer);
 }
 
+/**
+ * Handle the `/api/settings` HTTP endpoint (GET and POST).
+ *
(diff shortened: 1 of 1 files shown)


jniebuhr:
@cla-bot check

jniebuhr:
This doesn't do anything, settings->save(true); is already called in line 588
@crwper

crwper:
Shoot, you're right. The bug I was trying to fix was that Kff was not correctly updated on "Save and restart". But I think the bug was actually fixed by my other PR (splitting Kff out of the PID list), not by this one.

PR #696, opened 2026-05-12
Rewrite several loop methods to tasks

jniebuhr:
## Summary by CodeRabbit

* **Refactor**
  * Split periodic control into a dedicated logic loop and separate pacing to improve consistency of brew pressure/flow updates, progress tracking, and steam readiness.
  * Aligned background task pacing for steadier processing in standby vs active modes.
* **Bug Fixes**
  * Improved initialization sequencing to reduce missed connections when the display becomes ready.
* **Chores**
  * Tuned main-loop and DNS/network timing for more stable UI and background updates.


salekseev:
Nice split — moving the brew progression into `loopLogic` on core 0 directly addresses the original 5-10s phase overrun.

One related gap that survives this PR: `SmartGrindPlugin` still does a blocking `HTTPClient.GET()` synchronously from the `controller:grind:start` / `controller:grind:end` listeners (`src/display/plugins/SmartGrindPlugin.cpp:30-39`). With the new layout those listeners fire on `loopLogic` (core 0, prio 3). If the user's Tasmota relay is offline or on flaky WiFi, the default HTTPClient timeout (~5s) will stall `loopLogic` for the duration — which matches a reported reproduction signature (a phase running 5-10s long, occasionally). Worth fixing while the task layout is fresh.

Proposed fix: keep the event handlers as thin "intent" producers, run HTTP on a dedicated low-priority worker pinned to core 0. Latest-intent-wins via `xQueueOverwrite` so rapid toggles can't pile up stale commands; the `OFF -> ON` mode is modelled as a single compound action so it survives a queue overwrite atomically.

**`SmartGrindPlugin.h`**
```cpp
#ifndef SMARTGRINDPLUGIN_H
#define SMARTGRINDPLUGIN_H
#include "../core/Plugin.h"
#include <Arduino.h>
#include <freertos/queue.h>

const String COMMAND_ON = "Power%20On";
const String COMMAND_OFF = "Power%20off";

constexpr int SG_MODE_OFF = 0;
constexpr int SG_MODE_OFF_ON = 1;
constexpr int SG_MODE_ON_OFF = 2;

class SmartGrindPlugin : public Plugin {
  public:
    void setup(Controller *controller, PluginManager *pluginManager) override;
    void loop() override {};

  private:
    enum class Action : uint8_t { On, Off, OffThenOn };

    void start();
    void stop();
    void controlRelay(const String &command);

    static void workerTask(void *arg);

    Controller *controller = nullptr;
    QueueHandle_t actionQueue = nullptr;
};

#endif // SMARTGRINDPLUGIN_H
```

**`SmartGrindPlugin.cpp`**
```cpp
#include "SmartGrindPlugin.h"
#include <HTTPClient.h>
#include <WiFi.h>
#include <display/core/Controller.h>
#include <display/core/Event.h>

namespace {
constexpr UBaseType_t SMARTGRIND_TASK_PRIORITY = 1;     // below loopControl/loopLogic after this PR
constexpr BaseType_t  SMARTGRIND_TASK_CORE     = 0;     // co-located with WiFi / AsyncTCP
constexpr TickType_t  SMARTGRIND_OFF_ON_DELAY  = pdMS_TO_TICKS(500);
constexpr uint16_t    SMARTGRIND_HTTP_TIMEOUT  = 2000;  // bound worst-case worker time
}

void SmartGrindPlugin::setup(Controller *controller, PluginManager *pluginManager) {
    this->controller = controller;

    actionQueue = xQueueCreate(1, sizeof(Action));
    if (actionQueue == nullptr) {
        ESP_LOGE("SmartGrind", "queue alloc failed; plugin disabled");
        return;
    }
    xTaskCreatePinnedToCore(workerTask, "SmartGrind", configMINIMAL_STACK_SIZE * 8, this,
                            SMARTGRIND_TASK_PRIORITY, nullptr, SMARTGRIND_TASK_CORE);

    pluginManager->on("controller:grind:start", [this](Event const &) { start(); });
    pluginManager->on("controller:grind:end",   [this](Event const &) { stop();  });
}

void SmartGrindPlugin::start() {
    if (actionQueue == nullptr) return;
    if (controller->getSettings().getSmartGrindMode() != SG_MODE_ON_OFF) return;
    Action action = Action::On;
    xQueueOverwrite(actionQueue, &action);
}

void SmartGrindPlugin::stop() {
    if (actionQueue == nullptr) return;
    int mode = controller->getSettings().getSmartGrindMode();
    if (mode == SG_MODE_OFF) return;
    Action action = (mode == SG_MODE_OFF_ON) ? Action::OffThenOn : Action::Off;
    xQueueOverwrite(actionQueue, &action);
}

void SmartGrindPlugin::controlRelay(const String &command) {
    if (WiFi.status() != WL_CONNECTED) {
        ESP_LOGD("SmartGrind", "skipping relay (wifi status=%d)", WiFi.status());
        return;
    }
    HTTPClient http;
    String url = "http://" + controller->getSettings().getSmartGrindIp() + "/cm?cmnd=" + command;
    http.setConnectTimeout(SMARTGRIND_HTTP_TIMEOUT);
    http.setTimeout(SMARTGRIND_HTTP_TIMEOUT);
    http.begin(url);
    int code = http.GET();
    if (code != 200) {
        ESP_LOGW("SmartGrind", "relay request failed (code=%d)", code);
    }
    http.end();
}

void SmartGrindPlugin::workerTask(void *arg) {
    auto *self = static_cast<SmartGrindPlugin *>(arg);
    for (;;) {
        Action action;
        xQueueReceive(self->actionQueue, &action, portMAX_DELAY);
        switch (action) {
        case Action::On:        self->controlRelay(COMMAND_ON);  break;
        case Action::Off:       self->controlRelay(COMMAND_OFF); break;
        case Action::OffThenOn:
            self->controlRelay(COMMAND_OFF);
            vTaskDelay(SMARTGRIND_OFF_ON_DELAY);
            self->controlRelay(COMMAND_ON);
            break;
        }
    }
}
```

Why these specific choices:

- **Core 0** for the worker — same core as WiFi / AsyncTCP, where HTTP belongs.
- **Priority 1** — below `loopControl` (2) and `loopLogic` (3) once this PR lands; the worker can never preempt brew control even if blocked in TCP. (If merged before the task-split, priorities tie with the old layout — sequence the merges or fold this into the same PR.)
- **`xQueueOverwrite` with depth 1** — if the user mashes the button or the network drops, only the latest intended relay state matters; old commands are dropped.
- **`Action::OffThenOn` as a single compound action** — the existing `stop()` does OFF, `delay(500)`, ON for `SG_MODE_OFF_ON`. Modelling it as one queue entry preserves that semantic across a queue overwrite (the 500 ms wait moves into the worker via `vTaskDelay`, so the caller is no longer blocked).
- **Explicit 2 s HTTPClient timeouts** — bound worst-case worker time; without `setConnectTimeout` the LwIP default is ~10 s on an unreachable IP.
- **Stack `configMINIMAL_STACK_SIZE * 8`** — matches the heavier-stack convention used elsewhere (e.g. `ShotHistoryPlugin`'s rebuild task) and gives HTTPClient + WiFi headroom.
- **`ESP_LOGW/D/E` instead of `printf`** — matches the rest of the codebase and lets users filter by tag.

Happy to open this as its own PR if it's easier to review separately from the task rewrite.

src/display/core/Controller.cpp
```diff
@@ -91,7 +91,8 @@ void Controller::setup() {
     this->onScreenReady();
 
     updateLastAction();
-    xTaskCreatePinnedToCore(loopTask, "Controller::loopControl", configMINIMAL_STACK_SIZE * 6, this, 1, &taskHandle, 1);
+    xTaskCreatePinnedToCore(loopTask, "Controller::loopControl", configMINIMAL_STACK_SIZE * 6, this, 2, &taskHandle, 0);
```
salekseev, line 96: Nit / follow-up: with `loopControl` now correctly on core 0, the remaining asymmetry is the Arduino `loop()` itself. The Arduino-ESP32 framework default is `CONFIG_ARDUINO_RUNNING_CORE=1` (see `cores/esp32/main.cpp` — `xTaskCreateUniversal(loopTask, ..., CONFIG_ARDUINO_RUNNING_CORE)`), and this repo's `platformio.ini` does not override it, so `loop()` (and therefore every plugin's synchronous `loop()` call via `pluginManager->loop()`) still runs on core 1 alongside LVGL.

After this PR that is mostly harmless because the time-critical work moved into `loopLogic`, but it means a blocking plugin `loop()` (e.g. `WebUIPlugin`'s OTA check) can still freeze the UI render. Adding `-DCONFIG_ARDUINO_RUNNING_CORE=0` to `platformio.ini` would move the main loop to core 0 alongside `loopControl` / `loopLogic` / BLE / AsyncTCP and leave core 1 as a dedicated LVGL core — consistent with the stated goal of "only the UI on core 1".

Not blocking for this PR; flagging because it's the natural completion of the same re-arrangement.

PR #704, opened 2026-05-21
Move BLE notification handling into main loop context

crwper:
In the current firmware, when a DiFluid Microbalance is used and the shot is started with the rocker switch, the BLE interface hangs--the display can still be operated, but weight stops updating and shots can't be started until the display is reset. See, for example, this help request:

https://discord.com/channels/951416527721230336/1464770788144976048/1464800904233353440

This happens because the display receives a BLE notification from the controller, and inside the notification context it sends a write with response to the Microbalance. The BLE stack blocks waiting for the response, but the response is never processed because we have not returned from the BLE notification callback.

To fix this, rather than handling BLE notifications immediately, they are added to a queue which is then processed in the main loop context. With this change, the notification callback returns immediately, the notification is handled in the main loop, and the write with response is correctly handled since the BLE stack is no longer blocked.

## Summary by CodeRabbit

* **Refactor**
  * Bluetooth notifications now enqueue and are parsed later, then dispatched from the main loop for ordered, non-blocking processing.
  * Startup now validates the event queue and aborts initialization if it cannot be created.

* **Bug Fixes**
  * Disconnects are handled via the queued pipeline to ensure consistent cleanup and callbacks.
  * Queue overflow discards oldest events to preserve recent updates; excess enqueues are logged/dropped.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jniebuhr/gaggimate/pull/704?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/lib/NimBLEComm/src/NimBLEClientController.cpp
+++ b/lib/NimBLEComm/src/NimBLEClientController.cpp
@@ -5,6 +5,7 @@ constexpr size_t MAX_CONNECT_RETRIES = 3;
 
 NimBLEClientController::NimBLEClientController() : client(nullptr) {}
 
+// Create the event queue with the BLE client so callbacks can stay copy-only.
 void NimBLEClientController::initClient() {
     NimBLEDevice::init("GPBLC");
     NimBLEDevice::setPower(ESP_PWR_LVL_P9); // Set to maximum power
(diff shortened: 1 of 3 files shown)


crwper:
Looking into the two CodeRabbit comments...

crwper:
To address the CodeRabbit concerns:

1. Added a limit on the number of events handled in a single call to `dispatchPendingEvents`.
2. All disconnect actions are moved to the event handler so that all event actions occur in the order in which the events were received, e.g., disconnect doesn't null characteristics until after events have been replayed which occurred before disconnect.

crwper:
Let's hold off on this PR for a moment. I've just identified another place BLE writes occur within a BLE notification--starting with scale notifications--so it may be better to create a global BLE queue rather than one private to `NimBLEClientController`.

But also, the PR currently seems to be failing because of a permissions issue ("Deploy PR Preview / deploy (pull_request) Failing after 4m"). Are there any other errors with the current PR?

jniebuhr:
Superseded by #726

PR #712, opened 2026-05-26
feat(dashboard): Brew-mode quick-adjust steppers + over-pressure alert

FR4NKLN:
## Summary

Adds a Brew-mode quick-adjust workflow to the web Dashboard plus a visual cue when the boiler is still pressurised before a shot. Behaviour is consistent with the device's on-screen gear-icon BrewScreen Settings panel — no firmware semantics change beyond wiring up two missing WebSocket handlers.

### 1. Brew-mode +/- steppers (TEMP + WEIGHT/TIME)

When in Brew mode with no process running, the Process Controls card shows a TEMP stepper plus a second slot that morphs based on the firmware status flag `bt = isVolumetricAvailable && profile.isVolumetric`:

- `bt=true`  → label **TARGET WEIGHT**, value `Xg`, `+`/`−` adjusts the profile's volumetric target
- `bt=false` → label **TARGET TIME**, value `M:SS`, `+`/`−` adjusts the profile's phase duration

Both modes call the same `req:raise-brew-target` / `req:lower-brew-target` WS messages; `Controller::raiseBrewTarget` / `lowerBrewTarget` decide internally whether to adjust the profile's volumetric value or its phase durations based on the same condition. This is the exact logic the device's gear-icon BrewScreen Settings panel already uses — users get consistent affordances across both interfaces, and time-based profiles (or any brew without a connected scale) are no longer second-class citizens in the dashboard.

These are **transient "next-brew-only" adjustments**, not a profile editor. The firmware reloads the selected profile from disk after each brew completes (`Controller::deactivate`), so values revert. Permanent edits still go through the Profiles page.

Implemented in both `ProcessControls` (full layout) and `CompactProcessControls` (responsive layout, slimmer `MiniStepper` so two fit side-by-side).

### 2. Over-pressure red alert (idle only)

In Brew mode, the current pressure value turns `text-error` (red, semibold) when above 1.0 bar **and the brew is idle** — i.e. no shot active and no finished view shown. The cue is suppressed during the brew itself and the immediate post-brew finished view, since 8+ bar is the expected state during the shot and pressure takes a moment to bleed off after. The cue returns automatically once the user clears the finished view and the card is back to idle.

Visual purpose: warn the user that the boiler/group is still pressurised from a prior shot or heat-up before starting the next brew, without adding red noise during the periods when high pressure is normal.

Two reasons this matters:

- **Pressure-aware profiles depend on a clean baseline.** Adaptive / pressure-driven profiles read current pressure to drive phase transitions, predictive delay, and decline behaviour. Residual pressure above the profile's preinfusion thresholds at process start can skip or misadvance phases, producing a shot that doesn't follow the intended curve.
- **Back-pressurised group heads channel.** Common physical cause of blown pucks and uneven extraction, especially first shot of a session or after steaming.

Localised to the current pressure reading only (target + unit stay normal colour), and only triggers in Brew mode (>1 bar is normal in Steam).

### 3. Bonus fix: `potrait` typo in responsive variant

While working on the dashboard I noticed `Home/index.jsx` has `potrait:md:hidden` and `potrait:md:contents`. Tailwind silently ignores undefined variants, so the portrait switch between Compact and full layouts never worked — Compact was always shown in portrait orientation, even on tablets at widths where the full layout fits comfortably. Renamed to `portrait:` so the intended responsive behaviour activates.

### Firmware changes (minimal)

- `WebUIPlugin::handleWebSocketData`: add `req:raise-brew-target` / `req:lower-brew-target` handlers. The web client already sent these via the existing `raiseTarget()` callback (the Grind +/- buttons use the same pattern) but firmware had no matching handler — they were silently dropped.
- `Controller::deactivate`: reload the selected profile from disk on brew end, so stepper adjustments are transient.

## Test plan

- [ ] In Brew mode (idle), `−`/`+` adjusts TEMP — value updates immediately
- [ ] In Brew mode with BLE scale connected and a volumetric profile → second stepper labelled `TARGET WEIGHT`, `−`/`+` adjusts weight
- [ ] In Brew mode without scale (or with a non-volumetric profile) → second stepper labelled `TARGET TIME` (M:SS), `−`/`+` adjusts duration
- [ ] Disconnect/reconnect the scale mid-session → second stepper morphs between WEIGHT and TIME automatically on the next status tick
- [ ] Start a brew with adjusted values, let it finish — values revert to profile defaults afterward
- [ ] Pressure > 1 bar in Brew mode **while idle** → current value renders in red
- [ ] Pressure > 1 bar during an **active** brew or in the **finished** post-brew view → renders normal colour (red suppressed)
- [ ] Same > 1 bar value in Steam/Water/Standby/Grind → no red regardless of state
- [ ] Open dashboard in portrait orientation on a tablet ≥ 768px wide → full `ProcessControls` renders (not Compact)
- [ ] Compact view (narrow window / phone) shows the steppers side-by-side via `MiniStepper`

## Summary by CodeRabbit

* **New Features**
  * Brew target raise/lower controls now work from the UI.
  * Compact stepper controls for temperature and weight/time in brew idle.
  * Quick brew-mode controls added; time stepper displays M:SS format.

* **Bug Fixes**
  * Fixed responsive layout class name typo.

* **Improvements**
  * Pressure value shows alert styling when high during idle.
  * Profile values automatically revert to saved settings after each brew.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jniebuhr/gaggimate/pull/712?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/display/core/Controller.cpp
+++ b/src/display/core/Controller.cpp
@@ -651,8 +651,18 @@ void Controller::deactivate() {
         // raise/lowerBrewTarget) revert back to the profile's saved values
         // after each brew. For permanent changes the user should save via
         // the Profiles page or the display's Save button.
+        //
+        // Reset the profile in place before reload: parseProfile() appends to
+        // profile.phases rather than replacing, so feeding it the already-
+        // populated selectedProfile would double the phases on every brew
+        // (60s utility profile reading 1:00 → 2:00 → 3:00 across runs, and
+        // the brew loop actually executing the duplicated phases). Mirrors
+        // the `selectedProfile = Profile{}` pattern that
+        // ProfileManager::selectProfile already uses for the same reason.
         if (profileManager) {
-            profileManager->loadSelectedProfile(profileManager->getSelectedProfile());
+            Profile &p = profileManager->getSelectedProfile();
+            p = Profile{};
+            profileManager->loadSelectedProfile(p);
         }
     } else if (lastProcess->getType() == MODE_GRIND) {
         pluginManager->trigger("controller:grind:end");
--- a/src/display/core/Controller.cpp
+++ b/src/display/core/Controller.cpp
@@ -645,6 +645,15 @@ void Controller::deactivate() {
     currentProcess = nullptr;
     if (lastProcess->getType() == MODE_BREW) {
         pluginManager->trigger("controller:brew:end");
+        // Reload the selected profile from disk so any temporary temp/weight
+        // adjustments made via the dashboard +/- steppers or the display's
+        // gear-icon BrewScreen Settings panel (raise/lowerTemp,
+        // raise/lowerBrewTarget) revert back to the profile's saved values
+        // after each brew. For permanent changes the user should save via
+        // the Profiles page or the display's Save button.
+        if (profileManager) {
+            profileManager->loadSelectedProfile(profileManager->getSelectedProfile());
+        }
     } else if (lastProcess->getType() == MODE_GRIND) {
         pluginManager->trigger("controller:grind:end");
     }
--- a/src/display/plugins/WebUIPlugin.cpp
+++ b/src/display/plugins/WebUIPlugin.cpp
@@ -339,6 +339,13 @@ void WebUIPlugin::handleWebSocketData(AsyncWebSocket *server, AsyncWebSocketClie
                     controller->raiseGrindTarget();
                 } else if (msgType == "req:lower-grind-target") {
                     controller->lowerGrindTarget();
+                } else if (msgType == "req:raise-brew-target") {
+                    // Web dashboard Brew-mode +/- buttons. The web client has
+                    // long been sending this via raiseTarget() but no firmware
+                    // handler existed, so the message was silently dropped.
+                    controller->raiseBrewTarget();
+                } else if (msgType == "req:lower-brew-target") {
+                    controller->lowerBrewTarget();
                 } else if (msgType == "req:change-mode") {
                     if (doc["mode"].is<uint8_t>()) {
                         auto mode = doc["mode"].as<uint8_t>();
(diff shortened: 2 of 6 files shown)


FR4NKLN:
#### Why "TARGET" rather than just "weight" or "time"

The value is the setpoint the firmware is aiming for, not a guarantee of brew outcome. Under common conditions on certain profile shapes, the actual cup weight or brew time can land **dramatically** far from the value shown. Two concrete cases that come up in normal use:

- **TARGET WEIGHT 28g → actual cup 40g+.** Whether the firmware can guard against this depends on how the profile was authored. Profiles imported via the Gaggiuino, Decent TCL, or Meticulous JSON importers all carry a per-phase volumetric overflow cap by convention (Gaggiuino since PR #365 / v1.6.0; Decent and Meticulous since PRs #708 and #709 extended the same pattern). Hand-authored Gaggimate profiles vary — most guard only the final brew phase. On a profile of the latter shape, a low-resistance or channeled puck can dump 25-35g into the cup *before* the final-phase volumetric guard activates, and the brew sees the cup already past target on entry to the final phase.
- **TARGET TIME 1:40 → brew completes in 0:16.** Per-phase exit conditions (pressure >= N, flow >= N, pumped >= N) advance through phases much faster than their declared durations. A 100-second profile can resolve in 15-20 seconds when the pressure curve climbs fast against a free-flowing puck.

type: "standard" profiles (e.g., the bundled 9 Bar Espresso) are the structural exception: Phase::isFinished has an early-return at profile.h:109-111 that makes the volumetric target the only escape (besides a 5-minute safety cap). Standard profiles do stop at the target weight reliably.

**Worth flagging for the maintainer** concerning uniformity and intention

FR4NKLN:
@cla-bot check

PR #715, opened 2026-05-27
fix(firmware): close shot-history file/header inheritance race across…

FR4NKLN:
## Summary

When a user starts a second brew before the first brew's shot file has been finalized, `ShotHistoryPlugin` writes the new brew's samples into the previous brew's file under the previous brew's header. The user sees their most recent brew either missing from Shot History entirely, or labeled with the previous profile's name.

### What was wrong

`ShotHistoryPlugin::startRecording()` did not reset `isFileOpen` or `header`. The cleanup branch in `record()` only runs on its 250 ms tick AND only when `recording=false AND extendedRecording=false`. With a BLE scale connected, `extendedRecording` can stay `true` for up to 3 seconds after `controller:brew:end`.

The race:

1. Brew A ends. `endRecording()` flips `recording=false`, `extendedRecording=true`. Post-drip sampling continues. File still open with A's header.
2. User starts Brew B. `Controller::activate()` fires `controller:brew:clear` → `endExtendedRecording()` flips `extendedRecording=false`. Then `delay(200)` (shorter than the 250 ms `record()` interval, so the cleanup branch usually does NOT run). Then `controller:brew:start` → `startRecording()` flips `recording=true`.
3. Next `record()` tick sees `isFileOpen=true` so skips the file-open path. Brew B's samples get appended to A's file under A's header.
4. Cleanup eventually runs at end of B. Patches A's file header with B's `sampleCount` / `durationMs`. The index entry upsert at lines 622-628 overwrites the early-index entry (which had the correct "B" name) with the merged header.

Net effect: B's brew appears in Shot History under A's name and A's timestamp / id; B's separately-named `.slog` file is never created.

Compounded by a separate narrower race: `record()` reads `getSelectedProfile()` live on each tick to populate the header. If the user switches profile between `brew:start` firing and the first `record()` tick, the wrong name lands in the header even without the file-inheritance race.

### Fix

Two coupled changes:

1. **Extract the cleanup branch into `finalizeShotFile()`**, then call it defensively at the top of `startRecording()` when `isFileOpen=true`. The previous shot gets properly patched, closed, and indexed before the new shot's state is initialized.

2. **Populate the shot file header synchronously in `startRecording()`** — magic / sample-size / startEpoch / profileId / profileName / phaseTransitionCount — using the running `BrewProcess`'s per-shot immutable profile copy (`brewProcess->profile`). Closes the live-profile-read race. `record()`'s file-open path no longer initializes the header; `startRecording()` guarantees it.

The previously-dead `currentProfileName` field is repurposed as a mirror of the header value for downstream telemetry consistency.

### User-visible impact

Before: rapid back-to-back brews (within ~3s with BLE scale connected, or ~250 ms without) caused the second brew to be lost or mislabeled in Shot History.

After: each brew gets its own correctly-named entry in Shot History regardless of how quickly the next brew starts.

## Test plan

- [ ] Run brew A with profile A (BLE scale connected). Let it finish.
- [ ] Immediately start brew B with profile B (within a few seconds of A finishing, before A's extended recording would naturally complete).
- [ ] Open Shot History — both A and B appear with correct profile names in correct chronological order.
- [ ] Repeat without BLE scale connected — same expectation.
- [ ] Standard "wait a while between brews" workflow still works (no regression on the normal case).

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable shot recording: previous incomplete recordings are finalized before new ones start and very short/failed shots are discarded.
* **Improvements**
  * Profile details are captured at recording start for more accurate shot metadata and consistent history entries.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jniebuhr/gaggimate/pull/715?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/display/plugins/ShotHistoryPlugin.cpp
+++ b/src/display/plugins/ShotHistoryPlugin.cpp
@@ -101,22 +101,12 @@ void ShotHistoryPlugin::record() {
             currentFile = fs->open("/h/" + currentId + ".slog", FILE_WRITE);
             if (currentFile) {
                 isFileOpen = true;
-                // Prepare header
-                memset(&header, 0, sizeof(header));
-                header.magic = SHOT_LOG_MAGIC;
-                header.version = SHOT_LOG_VERSION;
-                header.reserved0 = (uint8_t)SHOT_LOG_SAMPLE_SIZE; // record sample size actually used
-                header.headerSize = SHOT_LOG_HEADER_SIZE;
-                header.sampleInterval = SHOT_LOG_SAMPLE_INTERVAL_MS;
-                header.fieldsMask = SHOT_LOG_FIELDS_MASK_ALL;
-                header.startEpoch = getTime();
-                Profile profile = controller->getProfileManager()->getSelectedProfile();
-                strncpy(header.profileId, profile.id.c_str(), sizeof(header.profileId) - 1);
-                header.profileId[sizeof(header.profileId) - 1] = '\0';
-                strncpy(header.profileName, profile.label.c_str(), sizeof(header.profileName) - 1);
-                header.profileName[sizeof(header.profileName) - 1] = '\0';
-                header.phaseTransitionCount = 0; // Initialize phase transition count
-                // Write header placeholder
+                // Header was populated synchronously in startRecording() —
+                // including profileId / profileName captured from the running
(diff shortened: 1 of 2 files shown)


jniebuhr:
@cla-bot check

PR #716, opened 2026-05-28
fix(firmware/web): /profiles page reliability + perf for large libraries

FR4NKLN:
## What & why

**Severity: high.** Once the favorited library is large — which happens fast,
since every profile is auto-favorited on import — both the `/profiles` page and
the on-device carousel become unreliable. Several failure modes converge around
30+ profiles: a `coex_enable` boot-loop, an `async_tcp` watchdog reboot when
`/profiles` loads, a long dashboard freeze during the periodic OTA check on
flaky networks, laggy selects after bulk saves, and a spinner-forever state on a
transient WS allocation failure. This PR scales both paths and addresses each.

### Firmware

- **Lazy-load the on-device carousel** (`DefaultUI`). Cold-start builds the
  favorited-ID list (cheap strings) but defers loading the `Profile` structs;
  only a small window around the displayed entry is resident at any time.
  Loading every favorited profile synchronously put 100 KB+ on the internal
  heap, starving the Wi-Fi/BLE coexistence allocator during Bluetooth bring-up
  (`coex_enable` aborts → boot loop). Lazy-load bounds the resident set
  regardless of library size.

- **O(1) incremental cache mutators** for select / favorite / unfavorite / save,
  replacing a blanket invalidation that re-read *every* favorited profile from
  SPIFFS on the profile-loop task after each change.

- **Newly-favorited profiles keep their configured order.** `onProfileFavorited`
  inserts at the carousel index derived from `getFavoritedProfiles()` (already
  `profileOrder`-sorted), discounting the selected entry pinned at index 0,
  instead of appending at the tail — so favoriting no longer reorders the
  carousel. (Addresses review feedback.)

- **Offload `req:profiles:list` to a dedicated worker task** (`WebUIPlugin`).
  Building a large list on the AsyncTCP task stalled the WebSocket and tripped
  the `async_tcp` task watchdog; it now runs on its own FreeRTOS task fed by a
  queue, so AsyncTCP returns immediately.

- **Null-check `ws.makeBuffer()`** at all four call sites — the buffer allocates
  from internal heap and can fail under fragmentation; results were previously
  dereferenced unconditionally.

- **Run the OTA update check on a one-shot task** so a slow HTTPS check can't
  block the main loop and stall the status broadcast.

### Web

- **`/profiles` load-error handling.** The list request is wrapped in try/catch;
  on failure the page shows an error alert with a Retry button instead of a
  permanent spinner, and non-array responses are handled defensively.

### Test plan

- [x] Boots cleanly on current `master` with a large auto-favorited library — no
      `coex_enable` abort; Wi-Fi, controller link (protocol match), and webserver
      all come up
- [x] Newly-favorited profile lands at its `profileOrder` slot in the carousel,
      not the tail
- [x] `/profiles` loads without an infinite spinner; a timeout / non-2xx surfaces
      the Retry banner
- [x] `req:profiles:list` build runs off the AsyncTCP task (worker queue) — no
      `async_tcp` watchdog reboot
- [x] Dashboard stays responsive while a background OTA check is slow
- [x] Rapid select / favorite / unfavorite is snappy (no O(N) SPIFFS reload per
      event)

### Notes
- The `ProfileManager::loadProfile` PSRAM-backing an earlier revision carried is
  dropped — #723 lands the same change.

## Summary by CodeRabbit

* **Performance**
  * Profile list and OTA checking now run off the main thread/task, improving responsiveness.
  * Profile cache switched to lazy loading with a bounded in-memory window for reduced memory use.

* **Improvements**
  * Added error messaging and a retry button when profile list loading fails.
  * WebSocket responses are now hardened to avoid crashes when message buffers cannot be allocated.

### DIFF
--- a/src/display/plugins/WebUIPlugin.cpp
+++ b/src/display/plugins/WebUIPlugin.cpp
@@ -176,7 +176,29 @@ void WebUIPlugin::loop() {
             }
         }
 
-        ws.textAll(statusDoc.as<String>());
+        // Per-client backpressure: skip clients whose send queue is already
+        // backed up so that 500 ms status ticks don't pile up behind a
+        // larger in-flight response (e.g. /profiles list build) and
+        // overflow the per-client message queue. PR #710 documented this
+        // guard in its description but the actual implementation never
+        // landed — only `setCloseClientOnQueueFull(false)` did, which
+        // prevented the close-on-overflow disconnect but didn't avoid the
+        // overflow drops. This is the missing half.
+        constexpr size_t STATUS_BACKLOG_LIMIT = 8;
+        const String payload = statusDoc.as<String>();
+        for (auto &client : ws.getClients()) {
+            if (client.status() != WS_CONNECTED) {
+                continue;
+            }
+            if (client.queueLen() >= STATUS_BACKLOG_LIMIT) {
+                // Client is slow or stuck behind an in-flight large
+                // response. Skip this tick; the next one is only 500 ms
+                // away and the dashboard will catch up once the backlog
+                // drains.
+                continue;
+            }
+            client.text(payload);
+        }
     }
     if (now > lastCleanup + CLEANUP_PERIOD) {
         lastCleanup = now;
@@ -361,16 +383,26 @@ void WebUIPlugin::handleWebSocketData(AsyncWebSocket *server, AsyncWebSocketClie
                     resp["msg"] = "Rebuild started";
                     size_t bufferSize = measureJson(resp);
                     auto *buffer = ws.makeBuffer(bufferSize);
-                    serializeJson(resp, buffer->get(), bufferSize);
-                    client->text(buffer);
+                    if (buffer != nullptr) {
+                        serializeJson(resp, buffer->get(), bufferSize);
+                        client->text(buffer);
+                    } else {
+                        ESP_LOGE("WebUIPlugin", "makeBuffer(%u) returned null for res:history:rebuild — client %u will see no response",
+                                 (unsigned)bufferSize, cid);
+                    }
                     ShotHistory.startAsyncRebuild();
                 } else if (msgType.startsWith("req:history")) {
                     JsonDocument resp(&psramAllocator);
                     ShotHistory.handleRequest(doc, resp);
                     size_t bufferSize = measureJson(resp);
                     auto *buffer = ws.makeBuffer(bufferSize);
-                    serializeJson(resp, buffer->get(), bufferSize);
-                    client->text(buffer);
+                    if (buffer != nullptr) {
+                        serializeJson(resp, buffer->get(), bufferSize);
+                        client->text(buffer);
+                    } else {
+                        ESP_LOGE("WebUIPlugin", "makeBuffer(%u) returned null for %s — client %u will see no response",
+                                 (unsigned)bufferSize, msgType.c_str(), cid);
+                    }
                 } else if (msgType == "req:flush:start") {
                     handleFlushStart(client->id(), doc);
                 }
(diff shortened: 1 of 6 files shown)


FR4NKLN:
@cla-bot check

src/display/plugins/WebUIPlugin.cpp
```diff
@@ -491,8 +527,78 @@ void WebUIPlugin::handleProfileRequest(uint32_t clientId, JsonDocument &request)
...
+
+void WebUIPlugin::profileListWorkerTask(void *arg) {
+    auto *plugin = static_cast<WebUIPlugin *>(arg);
+    ProfileListJob *job = nullptr;
+    while (true) {
+        if (xQueueReceive(plugin->profileListQueue, &job, portMAX_DELAY) == pdTRUE && job != nullptr) {
+            plugin->buildAndSendProfileList(*job);
+            delete job;
+            job = nullptr;
+        }
```
jniebuhr, line 543: shouldn't there be some delay? even just a small one?

src/display/ui/default/DefaultUI.cpp
```diff
@@ -281,14 +289,157 @@ void DefaultUI::loopProfiles() {
...
+        return;
+    }
+    if (std::find(favoritedProfileIds.begin(), favoritedProfileIds.end(), id) != favoritedProfileIds.end()) {
+        return; // already cached
+    }
+    // Append an empty placeholder. The Profile data is loaded lazily when
+    // the user navigates the carousel to this entry. Avoids a SPIFFS read
+    // on the event-firing task (usually AsyncTCP) for an entry the user
+    // may never look at.
+    favoritedProfileIds.emplace_back(id);
```
jniebuhr: This could change the order of profiles on the screen

src/display/plugins/WebUIPlugin.cpp
FR4NKLN: portMAX_DELAY blocks the task rather than polling — xQueueReceive suspends it
(Blocked state, zero CPU) until a job is posted and FreeRTOS wakes it on the
xQueueSend. So it's not a busy loop — a delay would only add latency before the
list builds. Standard blocking-consumer pattern — so no sleep needed here

src/display/ui/default/DefaultUI.cpp
FR4NKLN: Good catch — fixed. onProfileFavorited no longer appends at the tail; it derives
the insert index from the new favorite's position in getFavoritedProfiles()
(which is profileOrder sorted) — discounting the selected profile pinned at index
0 —  so the entry follows the configured order instead of jumping to the end.
Verified on-device: with a profile placed at profileOrder index 2 it inserted at
carousel index 2 rather than the tail (index 32 of 33). Pushed as 7e2358be.

src/display/ui/default/DefaultUI.cpp
```diff
@@ -281,14 +289,189 @@ void DefaultUI::loopProfiles() {
...
+            if (sorted[i] == selectedId) {
+                selectedBefore = 1;
+                break;
+            }
+        insertAt = static_cast<size_t>(1 + sortedPos - selectedBefore);
+        if (insertAt > favoritedProfileIds.size())
+            insertAt = favoritedProfileIds.size();
+    }
+    favoritedProfileIds.insert(favoritedProfileIds.begin() + insertAt, id);
+    favoritedProfiles.insert(favoritedProfiles.begin() + insertAt, Profile{});
```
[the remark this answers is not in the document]
FR4NKLN: Acknowledged — real but pre-existing rather than introduced here: the
incremental cache mutators already read/write favoritedProfileIds across the
event task and the render/profile-loop tasks without locking and the ordering
change doesn't add to that. Properly serializing the carousel (mutation +
render read + prefetch) is a broader change than this PR's scope, so I'm
deferring it as a separate follow-up rather than bolting partial locking on here.

src/display/ui/default/DefaultUI.cpp
```diff
@@ -281,14 +289,189 @@ void DefaultUI::loopProfiles() {
...
+        for (int i = 0; i < sortedPos; i++)
+            if (sorted[i] == selectedId) {
+                selectedBefore = 1;
+                break;
+            }
+        insertAt = static_cast<size_t>(1 + sortedPos - selectedBefore);
+        if (insertAt > favoritedProfileIds.size())
+            insertAt = favoritedProfileIds.size();
+    }
+    favoritedProfileIds.insert(favoritedProfileIds.begin() + insertAt, id);
```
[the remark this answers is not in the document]
FR4NKLN: Fixed in f28d6b8c — folded favoritedProfileIds.empty() into the early guard so
an empty cache (last favorite just removed) defers to the cold-start rebuild
instead of calling front() on an empty vector.

Issue #241, opened 2025-06-13
[Feature Request] Stabilize OTA


jniebuhr:
### Describe the feature

OTA currently is very dependent on connection and plugins enabled. We should try and download the firmware in the browser, then upload directly to screen and controller from there. We'll also just clear the web ui files and upload them as single files to keep profiles in the storage. This will require a firmware server to run somewhere

### Why is this feature useful?

Stabilize OTA process

### How complex is this feature?

High - Significant work required

PR #717, opened 2026-05-28
fix(firmware): mDNS responder reconnect leak — restore `.local` resolution

FR4NKLN:
**Severity: medium / chronic.** Long-standing community symptom: the
device is reachable by IP but `<hostname>.local` stops resolving after
the first WiFi flake / router rekey. Matches Issue #241 and several
closed PRs that took partial approaches.

## Root cause

`mDNSPlugin::setup` registered only a `controller:wifi:connect` event
listener. The listener called `MDNS.begin(hostname)` and added
services. There was no matching `MDNS.end()` anywhere — including on
disconnect — so every WiFi reconnect added a fresh layer of mDNS
state on top of the stale one from the previous association. The
underlying ESPmDNS responder ends up unable to authoritatively answer
A-record queries for `<hostname>.local`, even though the IP stays
reachable directly.

## Fix

- New `controller:wifi:disconnect` listener that calls `MDNS.end()`
  to tear down the responder cleanly when WiFi drops.
- Defensive `MDNS.end()` at the start of `start()` that runs if a
  prior responder is still tracked as active — covers the edge case
  where the disconnect event didn't fire (e.g. AP-fallback recovery
  paths where `WiFi.onEvent` registrations are gated on
  `WL_CONNECTED` and the disconnect handler was never installed).
- New `responderRunning` latch so end/begin can't double-fire on
  back-to-back events.

The hostname value itself is unchanged — `settings.getMdnsName()`
stays the single source of truth, no implicit suffixing or rewriting.
What's in Settings is what gets advertised.

## Test plan

- [x] Boot device, resolve `<hostname>.local` from a same-subnet
      client — works.
- [x] Force a WiFi reconnect (router rekey or `WiFi.disconnect()` +
      reconnect from serial console). Confirm `MDNS.end()` log line
      fires on disconnect, fresh `MDNS.begin()` log fires on reconnect.
- [x] Resolve `<hostname>.local` again — works.

## Summary by CodeRabbit

* **Bug Fixes**
  * The mDNS responder now properly stops and cleans up resources whenever WiFi connectivity is lost or interrupted.
  * Added defensive initialization procedures to ensure any previously running responders are completely shut down before attempting to start new instances.
  * Enhanced state management and validation to safely handle responder initialization and shutdown edge cases without errors.
  * Expanded diagnostic logging to provide better visibility into responder startup events and WiFi disconnection lifecycle.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jniebuhr/gaggimate/pull/717?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/display/plugins/mDNSPlugin.cpp
+++ b/src/display/plugins/mDNSPlugin.cpp
@@ -11,12 +11,27 @@ static constexpr char LOG_TAG[] = "mDNSPlugin";
 void mDNSPlugin::setup(Controller *controller, PluginManager *pluginManager) {
     this->controller = controller;
     pluginManager->on("controller:wifi:connect", [this](Event const &event) { start(event); });
+    pluginManager->on("controller:wifi:disconnect", [this](Event const &) { stop(); });
 }
-void mDNSPlugin::start(Event const &event) const {
+
+void mDNSPlugin::start(Event const &event) {
     const int apMode = event.getInt("AP");
     if (apMode)
         return;
-    if (!MDNS.begin(controller->getSettings().getMdnsName().c_str())) {
+
+    // Defensively tear down any prior responder before starting a new one.
+    // Without this, every WiFi reconnect leaks the previous mDNS records —
+    // the underlying ESPmDNS state ends up with stale entries and clients
+    // can no longer resolve `<hostname>.local` even though the device is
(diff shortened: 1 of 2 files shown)


FR4NKLN:
@cla-bot check

PR #718, opened 2026-05-28
fix(firmware/web): /scales page reliability + UX surfacing for BLE scale lifecycle

FR4NKLN:
Improves /scales reliability + UX: the scan spinner now tracks the actual scan window instead of clearing as soon as the POST returns, connected scales no longer vanish from the list when no scan has run, connection errors and scale disconnects are now surfaced to the user, and saved-scale auto-reconnect no longer races a user-initiated connect.

**Firmware**
- Auto-reconnect now gated on `!doConnect` so it can't override a pending user-initiated connect.
- `establishConnection` clears `doConnect` at entry (previously it was only cleared in `disconnect()`).
- Saved-scale retry now bounded (`SAVED_SCALE_RETRY_BUDGET = 60`); resets on find / scan / connect. Previously the auto-reconnect path ran every tick when the saved scale was powered off.
- New `scale:scan:complete` / `connect:error` (with reason code) / `connect:success` / `disconnect` events from `BLEScalePlugin`, re-broadcast by `WebUIPlugin` as `evt:scale:*`. Scan window time-boxed via `SCAN_DURATION_MS = 5000` to drive the `scan:complete` emission.
- Existing public `scale` accessors hardened to local-capture the raw pointer via `scale.get()` before deref — narrows the check-then-use race but doesn't eliminate UAF; a full fix (mutex / `shared_ptr` + `atomic_load`) belongs in a broader concurrency pass.
- `update()` split into `pollConnectedScale` / `tryAutoReconnectSaved` helpers (SonarQube complexity).

**Frontend**
- Spinner driven by `machine.scale.scanning` instead of a local flag; also cleared on WS close so a mid-scan socket drop doesn't leave it stuck.
- Connect-error and scale-disconnect banners — dismissible, 60s auto-clear, also cleared on standby or when the scale reconnects.
- `/scales` no longer hides a connected scale when the discovered list is empty.
- `fetch()` checks `response.ok` before parsing.

## Release Notes

* **New Features**
  * Added connection failure alerts with descriptive error messages
  * Added timed scale disconnection notifications that auto-dismiss

* **Bug Fixes**
  * Improved BLE scan completion with deadline-driven timing instead of polling
  * Enhanced connection and reconnection behavior for greater reliability
  * Fixed UI stuck in scanning state after connection loss
  * Strengthened error handling with structured feedback for connection failures

[Review Change Stack](https://app.coderabbit.ai/change-stack/jniebuhr/gaggimate/pull/718?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/display/plugins/BLEScalePlugin.h
+++ b/src/display/plugins/BLEScalePlugin.h
@@ -29,7 +29,6 @@ class BLEScalePlugin : public Plugin {
 
     void setup(Controller *controller, PluginManager *pluginManager) override;
     void loop() override;
-    ;
 
     void connect(const std::string &uuid);
     void scan();
--- a/src/display/plugins/BLEScalePlugin.h
+++ b/src/display/plugins/BLEScalePlugin.h
@@ -110,6 +110,13 @@ class BLEScalePlugin : public Plugin {
 
   private:
     void update();
+    // update() helpers — split out so the function stays under the
+    // cognitive-complexity / nesting limits enforced by SonarQube.
+    // `pollConnectedScale` runs when a scale pointer exists (whether
+    // currently connected or in mid-reconnect-retries); `tryAutoReconnect`
+    // runs when no scale pointer exists and a saved-scale address is set.
+    void pollConnectedScale(bool hasConnectedScale);
+    void tryAutoReconnectSaved();
     void onProcessStart() const;
     void pollScaleMetadata();
 
--- a/src/display/plugins/BLEScalePlugin.h
+++ b/src/display/plugins/BLEScalePlugin.h
@@ -8,6 +8,19 @@ void on_ble_measurement(float value);
 
 constexpr unsigned long UPDATE_INTERVAL_MS = 1000;
 constexpr unsigned int RECONNECTION_TRIES = 15;
+// Cap on how many consecutive update() ticks (at UPDATE_INTERVAL_MS each)
+// the saved-scale auto-reconnect path will run without seeing the saved
+// scale in the discovered list. After this we stop trying until a manual
+// scan / mode change / reboot resets the counter. Without this cap the
+// loop runs every 1 s indefinitely when the saved scale is powered off,
+// hammering the scanner and producing no value.
+constexpr unsigned int SAVED_SCALE_RETRY_BUDGET = 60;
+// How long the firmware considers an async BLE scan to be "in progress"
+// before emitting scale:scan:complete. NimBLE's `initializeAsyncScan` in
+// the remote-scales lib doesn't surface a completion callback, so we
+// time-box it. ~5 s matches the lib's internal default and is long
+// enough for sleeping scales to advertise at least once.
+constexpr unsigned long SCAN_DURATION_MS = 5000;
 
 class BLEScalePlugin : public Plugin {
   public:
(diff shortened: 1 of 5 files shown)


src/display/plugins/BLEScalePlugin.h
```diff
@@ -19,25 +32,39 @@ class BLEScalePlugin : public Plugin {
...
+    // use across `if (scale != nullptr) scale->foo()` is unsafe whenever
+    // another task (the bluetooth-disconnect event handler on core 0, or
+    // `update()`'s reconnection-tries timeout on core 1) can null `scale`
+    // between the check and the call. Mirror of the local-capture pattern
+    // already applied to `Controller::onVolumetricMeasurement` for the
+    // same family of cross-core UAF.
+    bool isConnected() {
+        auto *s = scale.get();
+        return s != nullptr && s->isConnected();
+    };
```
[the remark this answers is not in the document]
FR4NKLN: > Local raw-pointer capture is still unsafe across cores

Thanks — this is exactly the limitation called out in the PR description and
the corresponding commit message. The local-capture pattern is deliberate
harm-reduction (mirroring the same shape already used in
`Controller::updateControl` and PR #714's `onVolumetricMeasurement` fix), not
a claim of thread safety.

The complete fix — single mutex or `shared_ptr` + `atomic_load` — needs to
cover the analogous races in `Controller::deactivate` / status-broadcast read
of `selectedProfile.phases` / `Settings::doSave` that are already catalogued.
These should land those together in a dedicated concurrency-hardening PR rather
than partial coverage here.

FR4NKLN:
> Handle non-2xx responses for scan/connect actions

Done in dbb96ca7 — `response.ok` check added to both `onScan` and `onConnect`, errors flow through the existing catch handlers.

> Clear transient scan state when WebSocket closes

Done in ead53866 — `_onClose` now resets `scale.scanning` to false alongside the existing `connected: false` write so a mid-scan WS drop no longer leaves the spinner latched.

FR4NKLN:
@cla-bot check

lib/NimBLEComm/src/NimBLEClientController.cpp
```diff
@@ -26,17 +26,20 @@ void NimBLEClientController::scan() {
...
     scanner->setAdvertisedDeviceCallbacks(this, true);
-    // BLE and WiFi share the single 2.4 GHz radio on ESP32-S3. The previous
-    // settings (100ms window every 2s, ACTIVE scan with TX scan-requests) held
-    // the radio long enough to spike WiFi RTT into the 200–700 ms range,
-    // making the web UI feel unresponsive. Drop to 1.25% duty (50 / 4000) and
-    // passive scan (listen-only, no TX). BLE discovery is still functional —
-    // just slightly slower — and WiFi gets the radio back.
-    scanner->setInterval(4000);
-    scanner->setWindow(50);
+    // Active scan with 100 ms window every 2 s. The earlier 50 ms/4000 ms
```
jniebuhr: This is not where scales are being scanned. That scan is happening at a different location inside of the esp-arduino-ble-scales repo

jniebuhr (changes requested):

lib/NimBLEComm/src/NimBLEClientController.cpp
FR4NKLN: You're right — dropped 1b7094d4 and tightened the PR description.

PR #745, opened 2026-06-07
Fix pump power input clearing

### LINKED ISSUE #445: [Bug] [Web UI] Pump Power field cannot be emptied
### Brief summary of the bug

The field cannot be emptied because the field disappear and the Pump Mode is in fact unset

While it is somehow expected to have the Pump Mode set to off if setting the value to 0, it should be possible to delete the content for editing the value without having the field disappear.
It looks like a bug since the Pump Mode becomes unset and when saving the "flow" mode is selected.

At least in this case the Pump Mode should be set to Off **after saving** if an empty value is set.
IMHO, I would say that it would be better to have the control "on save" instead of "onchange" also for 0 value.

Yes

### Steps to reproduce

1. edit or create a profile
2. delete the pump power value

### Expected behavior

 the field remain visible and editable and the power mode set is still "Power"

Yes

```shell

```

### Firmware version

1.6.0

_No response_

### Additional context

_No response_

EyJunge1:
Fixes #445

## Summary
- Allow the Pump Power input to be cleared while editing
- Normalize empty or invalid values on blur and before save

## Test plan
- Edited a profile with Pump Mode set to Power
- Cleared the Pump Power field
- Entered a new value and verified it stays editable
- Ran `npm run lint:check` and `npm run build` in `web/`

## Summary by CodeRabbit

* **Bug Fixes**
  * Pump inputs now behave consistently across modes (power, pressure, flow), correctly handling empty or invalid text entries and avoiding unintended NaN/parsing issues.
  * Mode switches and pressure/flow edits reliably update displayed and persisted values.

* **New Features**
  * New pump input helpers improve live display, change handling, and on-blur normalization.
  * Pump power values are automatically normalized when saving profiles.


EyJunge1:
@coderabbitai resume

EyJunge1:
@cla-bot check

Issue #583, opened 2026-02-12
[Bug] Final weight is stuck at wrong value if the shot history is opened while the shot is running


tyumru:
### Brief summary of the bug

Profile history shows wrong weight value after entered in the middle of the shot

### Describe the issue in greater detail

I entered the profile history and tried to open the incomplete shot while the shot was still ongoing. The displayed weight value (between shot time and stars) stuck at 9.8g, even though the graph kept recording the correct values.

<img width="1218" height="799" alt="Image" src="https://github.com/user-attachments/assets/058a9f0a-dbe6-44af-98e9-1f1e456aeaf9" />

### Are you able to reproduce the issue?

Yes

### Steps to reproduce

1. power on the machine
2. connect bluetooth scale
3. start the shot
4. enter profile history
5. open the ongoing shot

### Expected behavior

I would expect the shot history shows the correct final weight

### Has this function/fetaure worked before with your GaggiMate?

Yes

### Relevant logs or error messages

```shell

```

### Firmware version

1.7.3

### What machine are you experiencing this issue on?

gaggia classic pro e24

PR #748, opened 2026-06-08
fix(display/core/plugins): Fixed issue #583 with the weight of the currently brewed shot being displayed incorrectly if shot history were to be opened

szymonmilewski:
## Summary

Fixes issue #583 
The ShotHistoryPlugin::handleRequest() method contains a loop, that updates contents of documents containing historical shots. Previously, even during brewing, if the shot history were to be opened, the current shot's weight would be displayed incorrectly. To address this, I added an additional check that verifies whether the current document being modified corresponds to the currently brewed shot. If so, and the current bluetooth wait is greater than 0, then update finalWeight which in the end updates o["volume"]. If one of those checks fail, then leave the finalWeight as it was. 

## Testing
 platformio run -e display succeeds

## Summary by CodeRabbit

* **Bug Fixes**
  * Shot history now shows live weight and volume for shots actively being recorded. Previously the history list could display outdated static values for the current recording; it now reflects accurate, real-time measurements so users see up-to-date information for active shots in the history view.

### DIFF
--- a/src/display/plugins/ShotHistoryPlugin.cpp
+++ b/src/display/plugins/ShotHistoryPlugin.cpp
@@ -465,7 +465,7 @@ void ShotHistoryPlugin::handleRequest(JsonDocument &request, JsonDocument &respo
                         o["samples"] = hdr.sampleCount;
                         o["duration"] = hdr.durationMs;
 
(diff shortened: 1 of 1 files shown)


jniebuhr:
Changed shot history logic for running shots not to appear

PR #753, opened 2026-06-10
Fix OTA controller update status after firmware update (#749)

### LINKED ISSUE #749: [Bug] system update web screen wrong behavior
### Brief summary of the bug

save and refresh race condition show there is an update (but there isnt)

updating to latest nigthly, going back to system update web screen the page showed that the controller is still on older version (like i havnt update) and that there is a new version to update (which I already updated to) this is the first but but I believe this was the behavior since the start and need to save and refresh what happened next is:
i clicked save and refresh (maybe even 2 times. the end result (see in picture) is that controller version shows correct updated version but there is also text saying I need to update to that same version.
this can be bypassed of course by on/off or maybe even move between pages and its a small bug but its still there.

Image

Yes

### Steps to reproduce

its probably about updating the controller and clicked save and refresh a few times after.
if its a race condition it will be hard to reproduce. better put claude on this he will find where this is :)

### Expected behavior

dont show controller is not updated. if its updated dont show it still needs update.

Yes

```shell

```

### Firmware version

latest nigthly see in pucture

GC

### Additional context

_No response_

EyJunge1:
## Summary
- Fixes #749: System Update screen no longer shows a controller update prompt for the version already installed
- Syncs the cached OTA controller version with the live `SystemInfo` version before computing `controllerUpdateAvailable`

## Test plan
- [x] `pio run -e display`
- [x] Verify controller version is correct and no update prompt appears when already on latest version

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed a duplicated OTA settings field in the Web UI response.
  * Prevented unnecessary controller version resync by updating the Web UI only when the controller version changes.

* **Refactor**
  * Simplified the OTA status refresh workflow: controller version is cached and reused, JSON response assembly is cleaned up, and both periodic and post-OTA refreshes follow the same streamlined update flow.


EyJunge1:
@cla-bot check

PR #783, opened 2026-06-19
fix(comms): deliver controller config reliably across BLE reconnects

centercirclesolutions:
## Deliver controller config reliably across BLE reconnects

**Problem.** When the controller reboots while the display stays up, the boiler often won't heat. The controller gets its PID/pressure/pump gains from the display as a config burst sent in reply to `SystemInfo` — which is exchanged *during* the BLE connect/subscribe handshake. The session reset that runs on connection events discards the in-flight burst, so the controller keeps firmware-default gains (`Kff 0.000`) and heater output stays 0. Only bites on a reconnect.

**Relationship to `a7ff7b3b` ("fix: Send pid settings on mode change").** That commit addresses the most visible symptom (Kff=0 → no heat) by re-pushing PID gains on every mode change. It covers PID only — pressure scale and pump coefficients are still not re-sent — and only after the user changes mode. This PR fixes the lost-burst race at its source so the full config burst arrives reliably on every reconnect, no user action required. The `setMode → setPidSettings` call from `a7ff7b3b` then becomes a harmless safety net; I've kept it for now and we can drop it in a follow-up if you prefer.

**Fix.**
1. `BleClientTransport` — emit (reset the session) **before** subscribing, so the config exchange lands in a clean session.
2. `Endpoint::handleConnection` — reset only on a real connect/disconnect transition, not on every redundant connect-emit (NimBLE fires `onConnect` + `onSubscribe`).
3. `GaggiMateServer` — announce `SystemInfo` only after the link is stable (debounce), then re-announce until PID settings arrive (convergence) so a single lost burst self-heals.

**Validation.**
- `move` lineage (re-order): **VERIFIED** — HW A/B, controller-reboot config lands every cycle.
- `master` (+ debounce): direction validated (0/4 → 3/4); convergence re-announce closes the remaining gap.
- Build: `pio run -e display -e controller` clean against current `master` (post-rebase).

Full write-up + current/corrected flow diagrams: `reports/ble-reconnect-config-delivery.html`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection state tracking so session reset logic only triggers on genuine connection-state transitions, reducing unintended resets.
  * Enhanced system information announcements by waiting for a stable link and retrying until configuration is received or the retry limit is reached.
  * Fixed the Bluetooth client connection flow to notify readiness before subscribing to TX notifications, with clearer handling of subscription failures.

### DIFF
--- a/lib/NanoPbComm/src/Endpoint.cpp
+++ b/lib/NanoPbComm/src/Endpoint.cpp
@@ -251,21 +251,27 @@ void Endpoint::handleData(const uint8_t *data, size_t length) {
 }
 
 void Endpoint::handleConnection(bool connected) {
-    lock();
-    _inFlight = false;
-    _retries = 0;
-    _txLen = 0;
-    _inFlightId = 0;
-    _lastRxId = 0;
-    _nextId = 1;
-    _rttValid = false; // latency is per-link; don't carry a stale estimate across reconnects
-    _smoothedRttMs = 0;
-    _lastRttMs = 0;
-    _queue.clear();
-    unlock();
+    // NimBLE emits "connected" more than once per link (onConnect + onSubscribe);
+    // reset the session only on a real transition, else a redundant emit's xQueueReset wipes an already-arrived burst.
+    const bool changed = connected != _connected;
+    _connected = connected;
+    if (changed) {
+        lock();
+        _inFlight = false;
+        _retries = 0;
+        _txLen = 0;
+        _inFlightId = 0;
+        _lastRxId = 0;
+        _nextId = 1;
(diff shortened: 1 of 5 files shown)


centercirclesolutions:
Superseded by #785, which fixes the same controller-reboot config-delivery issue more simply and completely. This PR mitigated it (~78%→~50% controller-reset fail) via session-reset reordering + a controller re-announce; #785 re-sends the config burst from the display for a few seconds after each reconnect and reaches 50/50 pass on bare master (no dependence on this PR's machinery). Closing in favor of #785.

PR #820, opened 2026-07-11
feat: add configurable water level indicator dot

hahny93:
**Summary**

Added a configurable water level indicator dot to the display, visible on menu/info/profile and brew/steam/water screens when a ToF/Alba water sensor is connected.

**Display:**
- 14px circle dot positioned at top center (Y=-230)
- Shows only when ToF sensor is present and the setting is enabled
- Uses sunriseIdle color when water level ≥20%, sunriseError color when low
- Both colors are configurable via existing Alba sunrise settings in the web UI

**Settings:**
- showWaterLevelDot toggle (default: on) in Settings → Machine → Alba Settings, below the External LED slider
- Persisted as NVS property "wld"


jniebuhr:
Replaced by the warnings feature

PR #821, opened 2026-07-12
fix(web): fix steam pump assist field snap back to previous value

khanguyen74:
- This PR fix the issue where Steam Pump assist field is not working properly; when the value changes, it keeps snapping back to the old value.
- Fix by introducing a temporary variable that holds the value while changing, and set the actual value on blur

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved steam pump percentage editing in Machine Settings.
  * Values now stay stable while typing, and are converted correctly when saved after leaving the field, based on the selected pressure unit.

### DIFF
--- a/web/src/pages/Settings/tabs/MachineTab.jsx
+++ b/web/src/pages/Settings/tabs/MachineTab.jsx
@@ -1,3 +1,4 @@
+import { useState } from 'preact/hooks';
 import { computed } from '@preact/signals';
 import { machine } from '../../../services/ApiService.js';
 import Section from '../../../components/Card.jsx';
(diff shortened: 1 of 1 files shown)


web/src/pages/Settings/tabs/MachineTab.jsx
```diff
@@ -216,13 +219,18 @@ export function MachineTab({ formData, onChange, setField }) {
               step='0.1'
               className='grow'
               placeholder={pressureAvailable.value ? '0.0' : '0.0 %'}
-              value={String(formData.steamPumpPercentage * (pressureAvailable.value ? 0.1 : 1))}
```
princeku318-commits, line 219: The pull request fixes this by introducing a local state (steamPumpDraft) on line 62:
And updating the input field handling to:

Divide instead of multiply (/ 10) to correctly match the display logic:

JavaScript code:
+ steamPumpDraft ?? String ( formData . steamPumpPercentage / ( pressureAvailable . value ? 10 : 1 ) );

Use an onChange handler to store the user's active typing state in steamPumpDraft so it doesn't snap back.

Reset the draft to null on onBlur only after the finalized value is committed via setField.

PR #836, opened 2026-07-22
Hardware Scales

dragm83:
- Add support for 2xHX711 Hardware scales
- Modify protocol to add scale value, scale calibration parameters
- Add related settings in Webui - user can select preferred scale source 
- Grind function always uses Bluetooth (with fallback to none/time)
- Add taring to Webui
- Modify controller volumetric source naming to be generic currentWeight / activeWeight / currentSource etc. - source selection in controller based on settings/nightly
- EEZ project is unmodified (kept systemStatus.bluetooth_scales which should be renamed in the EEZ project to something generic)

image

## Summary by CodeRabbit

* **New Features**
  * Added hardware-scale support with per-cell measurements, tare, calibration, and configurable sampling and filtering.
  * Added automatic selection between hardware and Bluetooth scales, with configurable preferred source.
  * Added web controls for scale selection, calibration, tare, and scale-factor editing.
  * Added scale diagnostics, active-source status, and capability information.

* **Bug Fixes**
  * Weight displays, shot history, and volumetric measurements now consistently use the active scale.
  * Improved handling of invalid, unstable, or temporarily unavailable readings.
  * Preserved active-scale information when viewing older shot records.

### DIFF
--- a/lib/GaggiMateController/src/GaggiMateController.cpp
+++ b/lib/GaggiMateController/src/GaggiMateController.cpp
@@ -256,6 +256,17 @@ void GaggiMateController::setup() {
 
 void GaggiMateController::loop() {
     unsigned long now = millis();
+
+    // Keep zero tracking and idle display quantization out of the measurement
+    // path whenever the brew valve or pump can be adding liquid to the cup.
+    if (hardwareScale != nullptr) {
+        const float pumpPower =
+            pump != nullptr && pump->getPumpPowerPtr() != nullptr ? *pump->getPumpPowerPtr() : 0.0f;
+        const bool scaleActivity =
+            (valve != nullptr && valve->getState()) || pumpPower > 0.01f;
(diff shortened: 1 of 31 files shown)


keramblock:
@dragm83 This is a really cool feature, and I’d genuinely love to see it merged.

I was going through the PR and noticed the agent’s comment about the fixed SCALE_PUBLICATION_INTERVAL_MS = 90. From what I understand, the 90 ms interval makes perfect sense for avoiding accidental 5 SPS behavior with a nominal 10 SPS HX711, but with the new 80 SPS mode it also seems to cap the published weight updates at roughly 11 Hz, even though the internal filtering runs much faster.

Do you think that could affect brew-by-weight, flow calculation, or predictive stopping in practice, or is ~11 Hz publication intentional for the control path as well?

P.S. I’m not a C++ developer, so apologies if I’m misunderstanding the code or raising a non-issue.

PR #873, opened 2026-08-16
Recover controller startup after BLE subscription race

Clinteastman:
## What changed

- keep the controller's existing `SystemInfo` send on BLE connection/subscription
- resend `SystemInfo` once after the first framed ping from the display
- serialize connection callbacks and inbound payload handlers on the same dispatch task so messages cannot cross reconnection boundaries
- prevent Start from activating a process while startup is incomplete or the controller is disconnected

## Root cause

The existing controller send happens synchronously from the BLE subscription callback. On affected startups, that notification can be emitted before the display's `subscribe()` call has fully returned and its notification path is ready. Because the framed reliability layer never receives that notification, it has nothing to acknowledge or retry; meanwhile ordinary telemetry works later and the display remains on `Starting` with controller version `0.0.0`.

The first framed ping is sent only after the display has completed subscription and emitted its connected state. Receiving that ping therefore provides the missing application-level handshake. The controller now resends `SystemInfo` once at that point, through the existing reliable Endpoint transport. This keeps the existing protocol and ACK/retry mechanism intact.

The earlier display-side subscription-order theory was disproved on hardware and that change has been removed.

## User impact

The display can complete controller initialization even when the initial subscription-time notification loses the race. The Start guard also prevents the tap-through recovery path from creating a process against a half-initialized BLE session.

## Hardware evidence

Before the fix, a diagnostic trace showed:

- BLE connected and notification subscription completed
- live temperature telemetry continued
- no `SystemInfo` frame reached the display
- the display compared controller version `0.0.0` and stayed on `Starting`

The handshake fix was then built and installed on a GaggiMate Pro Rev 1.1 controller using the existing display-to-controller BLE OTA path. After hardware validation, review identified a cross-task reconnect edge case; the final commit now serializes connection callbacks and payload handlers through the existing dispatch task.

After the fix:

- the controller installed and rebooted successfully
- the display reported the updated controller build and protocol 3
- six consecutive display restarts all completed startup
- the tests covered both observed orderings: `SystemInfo` before the display's connected log and a delayed post-connect `SystemInfo` delivered by the first-ping resend
- the full production web UI was restored and verified after removing the temporary OTA test bridge

## Validation

- final controller firmware build passed: 11.7% RAM, 20.7% flash
- final display firmware build with the production web UI passed: 28.4% RAM, 66.5% flash
- live web UI returned HTTP 200 and live controller temperature after the clean flash
- `git diff --check` passed
- complete PR diff reviewed: 5 files; the final protocol dispatcher change addresses the review-identified reconnect race
- temporary diagnostic logging and local OTA test code are not present in the PR

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved controller activation reliability by preventing startup until initialization is complete, the controller is connected and ready, and no other process is active.
  * Improved post-connection synchronization by reliably delivering system information after the communication handshake and first protocol ping.
  * Prevented messages from a previous connection from being processed after reconnecting.
  * Improved connection and message event handling for more consistent communication behavior.

### DIFF
--- a/.clabot
+++ b/.clabot
@@ -35,7 +35,8 @@
     "szymonmilewski",
     "stephenwilley",
     "Marbit911",
-    "khanguyen74"
+    "khanguyen74",
+    "Clinteastman"
   ],
   "message": "We require contributors to sign our Contributor License Agreement, and we don't have yours on file. In order for us to review and merge your code, please contact @jniebuhr (mdwasp) on Discord to get yourself added."
 }
--- a/lib/NanoPbComm/src/Endpoint.cpp
+++ b/lib/NanoPbComm/src/Endpoint.cpp
@@ -8,7 +8,7 @@ static const char *ENDPOINT_TAG = "Endpoint";
 
 Endpoint::Endpoint(Transport &transport) : _transport(transport) {
     _mutex = xSemaphoreCreateRecursiveMutex();
-    _rxQueue = xQueueCreate(RX_QUEUE_DEPTH, sizeof(gm::Payload));
+    _rxQueue = xQueueCreate(RX_QUEUE_DEPTH, sizeof(DispatchEvent));
     if (_mutex == nullptr || _rxQueue == nullptr)
         ESP_LOGE(ENDPOINT_TAG, "Failed to allocate endpoint resources (out of memory)");
 }
@@ -42,10 +42,16 @@ void Endpoint::begin() {
 
 void Endpoint::dispatchTaskFn(void *arg) {
     auto *self = static_cast<Endpoint *>(arg);
-    gm::Payload payload;
+    DispatchEvent event;
     for (;;) {
-        if (xQueueReceive(self->_rxQueue, &payload, portMAX_DELAY) == pdTRUE)
-            self->dispatch(payload);
+        if (xQueueReceive(self->_rxQueue, &event, portMAX_DELAY) != pdTRUE)
+            continue;
+        if (event.isConnection) {
+            if (self->_connHandler)
+                self->_connHandler(event.connected);
+        } else {
+            self->dispatch(event.payload);
+        }
     }
 }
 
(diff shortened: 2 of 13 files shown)


Clinteastman:
@coderabbitai review

Clinteastman:
> We require contributors to sign our Contributor License Agreement, and we don't have yours on file. In order for us to review and merge your code, please contact @jniebuhr (mdwasp) on Discord to get yourself added.

Yeah I signed it and sent it back to @jniebuhr

Clinteastman:
@coderabbitai review

Clinteastman:
You are right: SystemInfo is already sent through the reliable endpoint and should be ACKed/retried. Looking again at the client ordering, I think the failure is narrower than the original PR wording suggested.

The controller's `onSubscribe()` can push SystemInfo while the display is still inside `_notifyChar->subscribe(...)`. The display may receive it, queue it and send the ACK immediately. When `subscribe()` then returns, `BleClientTransport::connectToServer()` calls `emitConnection(true)`, and `Endpoint::handleConnection(true)` resets `_rxQueue`. That can discard the already-ACKed SystemInfo before the dispatch task invokes `onSystemInfo()`. The controller correctly sees the ACK and therefore has nothing left to resend, while the display stays at `Starting`.

So I do not think the controller reliability mechanism is generally broken. The INFO read in this PR is an idempotent display-side recovery for the observed state. I have corrected the PR root-cause description accordingly. The direct architectural fix would be to make the client endpoint reset happen before notification subscription, but I kept this PR on the display-side path that was built and exercised against the official controller firmware: five consecutive restarts recovered cleanly.

Clinteastman:
Reworked this draft following the protocol discussion. The legacy INFO fallback has been removed.

The patch now fixes the reliable-delivery path directly by resetting the display endpoint before notification subscription. That closes the race where `SystemInfo` could be delivered, queued, and ACKed synchronously during `subscribe()`, then erased by the display's later connected-state reset. The controller would quite correctly stop retrying after that ACK.

Both display and controller firmware builds pass, and the complete upstream diff is now only two files. I have kept the PR in draft because this revised ordering fix has not yet had the physical restart test that the earlier fallback received.

Clinteastman:
Final diagnosis after hardware tracing:

The subscription-time SystemInfo notification can be emitted before the display notification path is fully usable. In that case it never reaches the framed Endpoint layer, so there is nothing for that layer to ACK or retry. Telemetry can begin later while the display still has controller version 0.0.0 and remains on Starting.

The final fix keeps the existing connect-time send and reliable transport. The first framed ping is used as the application-level readiness handshake, and the controller resends SystemInfo once through that same reliable Endpoint path. Connection changes and incoming payloads are also serialized through the same dispatch task so stale payloads cannot cross reconnect boundaries.

This supersedes the earlier INFO-fallback and client subscription-order explanations above. Both were tested and are not present in the final diff. The handshake behaviour passed six consecutive hardware restarts, including both observed event orderings. The subsequent dispatch-serialization hardening passes both firmware builds and the final automated review.

jniebuhr:
@cla-bot check

Issue #839, opened 2026-07-27
[Bug] Puck flow & puck resistance estimator only arms on the first shot(s) after power-on, then stays zero (no BLE scale)


vasto7:
> **Note: this machine has NO Bluetooth scale connected.** The "puck flow" discussed here is the firmware's pump-model / pressure-based estimate, not a scale-derived measurement. This rules out any scale connection/disconnection cause.

## Summary

Calculated **puck flow** and **puck resistance** stop being estimated after the first shot (or first couple of shots) following a power-on. Once they go to zero, they stay zero for every subsequent shot in that power session, and later power cycles do **not** reliably restore them.

Both values are clearly a single coupled estimator output: in the shot logs they appear together (both non-zero) or not at all (both zero) for the whole shot — they are never present independently.

## Environment

- Firmware: **v1.8.1** (controller and display)
- Hardware: **GaggiMate Pro Rev 1.1**, Stable channel
- Machine: **DeLonghi ECP 3420** (pump machine)
- **No Bluetooth scale** — puck flow here is the pump-model / pressure-based estimate, not scale-derived
- Boiler Refill plugin present
- Pump Flow Coefficients 10.205 / 5.521; Pressure Sensor Rating 16 bar

## Symptom

- Shots **26 and earlier**: calculated puck flow present.
- Shots **29 onward**: calculated puck flow zero.
- Puck flow (`pf`) and puck resistance (`pr`) are always zero together or non-zero together across the whole shot.
- In working shots, `pf`/`pr` arm once pressure stabilizes (~10 bar) and flow is ~3.5 ml/s, with puck resistance ramping up over the shot.
- In failing shots, the pressure (~10 bar), flow (~3.5 ml/s), and temperature curves are **near-identical** to the working shots on the **same profile**, yet the estimator never arms.

## Reproduction / test matrix

A single power session, all first shots (rarely any back-to-back shots):

| Action before shot | Boiler Refill | Shot | Puck flow calculated? |
|---|---|---|---|
| 1st power on | On | 40 | **Yes** |
|  | On | 41 | No |
|  | On | 42 | No |
|  | On | 43 | **Yes** |
| Manual flush | On | 44 | No |
| Manual flush | On | 45 | No |
| Turn off Boiler Refill, power cycle | Off | 46 | No |
| Manual flush | Off | 47 | No |
| Power cycle, manual flush | Off | 48 | No |
| Save and restart settings | Off | 49 | No |
[...]

PR #876, opened 2026-08-21
fix: Reset puck-flow filter accumulators on tare/reset (fixes part of #839)

b3nj1:
## Problem

#839: calculated puck flow (`pf`) / puck resistance (`pr`) work on the first
shot after power-on, then read zero on every shot after that.

## Cause

`PressureController::tare()`/`reset()` reset the puck-flow arming state
machine, but not the two IIR-filtered accumulators that feed it
(`_filteredPressureDerivative`, `_waterThroughPuckFlowRate`) or
`_lastPuckConductance`/`_puckConductance`. Their values from the end of one
shot (a sharp pressure-drop transient at group-valve-close) leak into the
next shot and can mask the signature the estimator looks for, so it never
arms.

## Fix

Reset those fields too, in both `tare()` and `reset()`.

## Test

Added `test/test_puckflow_latch` (native, `pio test -e native_puckflow`,
modeled on `native_autotune`): a two-shot scenario separated by
`tare()`+`reset()`. Fails on `master`, passes with the fix.

## Note

There's a separate, likely-related issue where `NimBLEClientController::tare()`
is a fire-and-forget BLE write with no delivery confirmation — if it's
dropped, the controller-side reset never runs at all. Not fixed here;
flagging as a possible follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pressure and puck-flow estimator behavior when starting a new brew shot after tare or reset.
  * Ensured estimator state is correctly reinitialized between successive shots.

* **Tests**
  * Added native automated coverage for pressure ramping, holding, depressurization, and repeated tare/reset cycles.
  * Added validation that puck-flow estimation arms correctly on the first shot and subsequent shots.

### DIFF
--- a/lib/GaggiMateController/src/peripherals/DimmedPump.cpp
+++ b/lib/GaggiMateController/src/peripherals/DimmedPump.cpp
@@ -45,7 +45,10 @@ float DimmedPump::getPuckFlow() { return _pressureController.getCoffeeFlowRate()
 
 float DimmedPump::getPuckResistance() { return _pressureController.getPuckResistance(); }
 
-void DimmedPump::tare() { _pressureController.tare(); }
+void DimmedPump::tare() {
+    _pressureController.tare();
+    _pressureController.reset();
+}
 
 void DimmedPump::loopTask(void *arg) {
     auto *pump = static_cast<DimmedPump *>(arg);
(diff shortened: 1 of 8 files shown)


b3nj1:
> We require contributors to sign our Contributor License Agreement, and we don't have yours on file. In order for us to review and merge your code, please contact @jniebuhr (mdwasp) on Discord to get yourself added.

@jniebuhr , I requested the license agreement here: https://discord.com/channels/951416527721230336/1342186457317769256/1540201338355327087

PR #908, opened 2026-09-13
Add support for dump valve on SSR2

fj-sanchez:
## Summary
- Adds an SSR2 **Dump valve** mode for machines with a separate group dump solenoid (Sage/Breville Barista Express). The 3-way stays on RELAY; SSR2/ALT drives the dump valve.
- After a shot or abort, the pump stops and the dump valve opens **while the 3-way stays open**, then both close. That relieves pump, thermoblock, and puck pressure through the dump path (drier pucks). Duration is configurable (default 2 s). A second stop during dump skips the rest.
- Profiles gain an optional per-phase `alt` field (Closed/Open in the editor when Dump valve is selected) so a backflush routine can open the dump solenoid between pressurize cycles.
- Ships `[Utility] Backflush (Dump valve)`. Classic `[Utility] Backflush` is unchanged. Gaggia/grind-on-SSR2 behavior is unchanged unless this mode is selected.

`deactivate(force)`: default stop starts dump; `force=true` tears down immediately (errors, leaving brew, mode change).

## Test plan
- [x] Machine tab: select **Dump valve**, set duration, save/reload
- [x] Profile editor shows **Dump valve** Closed/Open only in that mode; save/reload a phase with `alt: 1`
- [x] Brew with Dump valve: end of shot keeps 3-way on, pulses SSR2, then both off
- [x] Stop mid-shot: same dump-then-close sequence
- [x] Second stop during dump cuts it short
- [x] Flush button also dumps after Flush + Drain
- [x] SSR2 = Grind/None: no dump pulse, grind still works
- [x] Run `[Utility] Backflush (Dump valve)`: depressurize opens dump with 3-way still open

## Summary by CodeRabbit

* **New Features**
  * Added dump-valve support for compatible machines, including configurable relay behavior and valve-open duration.
  * Added a new “Backflush (Dump valve)” utility profile.
  * Added dump-valve controls to profile phase editors when enabled.
  * Brew completion and cancellation can now run a timed dump phase.
  * Added a “Dump valve” SSR2 function option and 0–10 second duration setting.
* **Bug Fixes**
  * Improved process handling so cleanup does not interrupt an active dump cycle.


fj-sanchez:
I have tested it with the simulator, and all seems to be working correctly. I will try it on my machine next.

